PR #1, opened 2018-02-19
Use constants from libc where they exist

faern:
Hi. I'm writing a wrapper for `libnftnl` in Rust to be able to control the Linux firewall. To do that I need to talk netlink. Since `libnftnl` uses `libmnl` for that natively I could just write a wrapper for `libmnl` as well. But the more pure-rust I can use the better was my thought. And it also feels like there are already too many libraries duplicating the `NlFamily` enum. So I looked for pure Rust netlink libraries.

After looking at yours for a while I see that many of the constants you declare are already in libc. So I thought I could try to make it use them instead. I don't know if you want this change or not, but it reduces duplication at least. We could quite likely get the remaining constants merged into libc as well, I have had constants from for example `<linux/netfilter/*>` merged before.

I had to do a little trick with the macro there to get rid of unnecessary `unsafe` warnings since there are now both safe and unsafe usages of that macro. But it turned out quite OK. Can be simplified again if all constants become safe as well.

### DIFF
--- a/src/c/netlink.c
+++ b/src/c/netlink.c
@@ -5,56 +5,6 @@
 
 const uint8_t nla_alignto = NLA_ALIGNTO;
 
-const uint32_t netlink_route = NETLINK_ROUTE;
-const uint32_t netlink_unused = NETLINK_UNUSED;
-const uint32_t netlink_usersock = NETLINK_USERSOCK;
-const uint32_t netlink_firewall = NETLINK_FIREWALL;
-const uint32_t netlink_sock_diag = NETLINK_SOCK_DIAG;
-const uint32_t netlink_nflog = NETLINK_NFLOG;
-const uint32_t netlink_xfrm = NETLINK_XFRM;
-const uint32_t netlink_selinux = NETLINK_SELINUX;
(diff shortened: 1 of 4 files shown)


jbaublitz:
@faern This seems to make a lot of sense for now. I will work on getting the rest merged into libc. Given that I'm already using libc for socket code, this would require no additional deps so I'm inclined to eliminate unnecessary C deps as well. I will see if I can get netlink constants merged into libc soon. If it is merged quickly, I will make some edits and switch them all over to libc. If it takes them a bit to get to my PR, I'll review and merge then switch over to libc once my PR is accepted. Thanks for the suggestion and PR!

faern:
Nice! Awesome crate by the way. Would be really nice to have comprehensive Netlink support in Pure rust without too many dependencies.

jbaublitz:
@faern rust-lang/libc#921 is now open. After looking at your proposed changes, I have a few more changes that I'd like to make on top of yours so I think I will merge and push some more changes then put out the next release on crates.io. I'll keep checking on that issue and will ping you once that process gets moving so we can hopefully get that C dependency out in future releases. Thanks again!

PR #3, opened 2018-02-19
Leave allocation up to the consumer of the library

faern:
Here is a suggestion for an API change.

For reading data into buffers it's quite customary to leave the allocation up to the consumer of the library. This does not force any allocation method on them and they can freely choose how to allocate the memory, if it should be on the stack or the heap etc. So as a suggestion to move towards that I propose this change.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -52,28 +52,21 @@ impl NlSocket {
 
     /// Send message encoded as byte slice to the netlink ID specified in the netlink header
     /// (`nl::nlhdr::NlHdr`).
-    pub fn send(&mut self, buf: &[u8], flags: i32) -> Result<isize, io::Error> {
+    pub fn send(&mut self, buf: &[u8], flags: i32) -> Result<usize, io::Error> {
         match unsafe {
             libc::send(self.fd, buf as *const _ as *const c_void, buf.len(), flags)
         } {
-            i if i >= 0 => Ok(i),
+            i if i >= 0 => Ok(i as usize),
             _ => Err(io::Error::last_os_error()),
         }
     }
 
     /// Receive message encoded as byte slice from the netlink socket.
-    pub fn recv(&mut self, len: Option<usize>, flags: i32) -> Result<Vec<u8>, io::Error> {
-        let mut v = match len {
-            Some(l) => vec![0; l],
-            None => Vec::new(),
-        };
+    pub fn recv(&mut self, buf: &mut [u8], flags: i32) -> Result<usize, io::Error> {
         match unsafe {
-            libc::recv(self.fd, v.as_mut_slice() as *mut _ as *mut c_void, v.len(), flags)
+            libc::recv(self.fd, buf as *mut _ as *mut c_void, buf.len(), flags)
         } {
-            i if i >= 0 => {
-                v.truncate(i as usize);
-                Ok(v)
-            },
+            i if i >= 0 => Ok(i as usize),
             _ => Err(io::Error::last_os_error()),
         }
     }
@@ -88,7 +81,7 @@ impl NlSocket {
 
     /// Serialize and send Rust `NlMsg` type
     pub fn sendmsg<I: Nl, T: Nl>(&mut self, mut msg: NlHdr<I, T>, flags: i32)
-                                        -> Result<isize, NlError> {
+                                        -> Result<usize, NlError> {
         let mut state = NlSerState::new();
         try!(msg.serialize(&mut state));
         let len = try!(self.send(state.into_inner().as_slice(), flags));
@@ -96,10 +89,11 @@ impl NlSocket {
     }
 
     /// Receive and deserialize Rust `NlMsg` type
-    pub fn recvmsg<I: Nl, T: Nl>(&mut self, len: Option<usize>, flags: i32)
+    pub fn recvmsg<I: Nl, T: Nl>(&mut self, len: usize, flags: i32)
                                        -> Result<NlHdr<I, T>, NlError> {
-        let mut buf = try!(self.recv(len, flags));
-        let msg = try!(<NlHdr<I, T> as Nl>::deserialize(&mut NlDeState::new(buf.as_mut_slice())));
+        let mut buf = vec![0; len];
+        let bytes_read = try!(self.recv(&mut buf, flags));
+        let msg = try!(<NlHdr<I, T> as Nl>::deserialize(&mut NlDeState::new(&buf[..bytes_read])));
         Ok(msg)
     }
 }


src/socket.rs
```diff
@@ -52,28 +52,21 @@ impl NlSocket {
...
+            i if i >= 0 => Ok(i as usize),
             _ => Err(io::Error::last_os_error()),
         }
     }
 
     /// Receive message encoded as byte slice from the netlink socket.
-    pub fn recv(&mut self, len: Option<usize>, flags: i32) -> Result<Vec<u8>, io::Error> {
-        let mut v = match len {
-            Some(l) => vec![0; l],
-            None => Vec::new(),
```
faern: This part was likely a bug unless I'm missing the use-case? `Vec::new` will return an empty vector and thus `v.as_mut_slice()` will be `[]` and have length zero. Thus it can never read any byte into that slice if `None` is passed for `len`.

src/socket.rs
```diff
@@ -88,18 +81,18 @@ impl NlSocket {
...
+        try!(self.send(state.into_inner().as_slice(), flags))
     }
 
     /// Receive and deserialize Rust `NlMsg` type
-    pub fn recvmsg<I: Nl, T: Nl>(&mut self, len: Option<usize>, flags: i32)
+    pub fn recvmsg<I: Nl, T: Nl>(&mut self, len: usize, flags: i32)
                                        -> Result<NlHdr<I, T>, NlError> {
-        let mut buf = try!(self.recv(len, flags));
-        let msg = try!(<NlHdr<I, T> as Nl>::deserialize(&mut NlDeState::new(buf.as_mut_slice())));
+        let mut buf = vec![0; len];
```
faern: Here is some internal allocation left. I didn't want to change too much before getting some feedback from you what you think of this type of change.

faern: For such a convenience function it might be fully OK to do internal allocation. As long as there is a way to use the library without it. That is separate, easy to use, `recv` and `deserialize`.

Problem here is that `buf` is dropped after each call. So it's quite expensive in the current state. Having to do a heap allocation for each read packet.

jbaublitz:
I will take a look at this soon but I agree that this makes more sense in retrospect given the low-level nature of what this library is meant to do. Feel free to make changes and I'll review them once the API is good to go and  the tests pass.

src/socket.rs
jbaublitz: Yes that seems to be a bug. If the user is required to allocate it from the application, that should remove the need for the size parameter as long as the proposed change is a fat pointer or something similar. One thing I will warn you about with netlink however is that if I recall correctly, netlink will drop any piece of the message that is not captured in a single read (the original reason I decided to use a `Vec`). Just keep that in mind when you make changes for the API as that's something that could definitely cause issues. This, however, might just be a documentation problem as opposed to something that needs to be guarded against in the code.

faern: This does not change how much a user can read. The length of the buffer is still fully controllable by the user.

faern:
Given that the length field is a `u32` we can't just allocate for the max message size, as it would be 4GiB (?). So the length of the buffer must be controllable by the user. On the other hand it's a bit off to require the user to send in a buffer but also give out a deserialized object (the user would still have the raw data in the buffer as well).

My thought is that maybe `recvmsg` is not needed. A user of the library could just chain `recv` and deserialization manually quite easily. But yes, if we keep this API then I think just giving the length and letting it allocate internally, like now, is fine.

src/socket.rs
jbaublitz: Just to clarify, by issues I mean for the user, not from a technical standpoint. I'm happy to use a fat pointer but I am more weighing options in my head here with regard to where the safeguard should be here: in the docs warning about data loss with improperly sized buffers or in the code and the more I think about it I tend to agree that more control here is probably better.

faern: Yes. Makes sense to say in the documentation that netlink will discard any part of a message that does not fit in the buffer. Linux does the same thing for UDP as well (completely unrelated).

jbaublitz:
@faern Ok let's take another look at this. I agree that it would be a bit clunky to deserialize while also leaving a buffer in.

My suggestion based on [the following](https://git.kernel.org/pub/scm/linux/kernel/git/davem/net-next.git/commit/?id=d35c99ff77ecb2eb239731b799386f3b3637a31e) is what I will probably end up doing: allocating the fairly negligeable 32k max for netlink messages on the stack and reworking this to be released when it is not actively serializing or deserializing. This can remove much of the usage of the heap in neli for the next release. After that, I will bump to 0.2.0 and upload to crates if there are no other concerns or suggestions at this time.

faern:
@jbaublitz 32k sounds small? I'm not that familiar with netlink. But I know the netfilter library `libnftnl` can have maximum messages of just above 64k: https://git.netfilter.org/nftables/tree/src/mnl.c#n44

jbaublitz:
@faern I haven't forgotten about this. A couple notes:

If I'm reading the kernel code vs the comment correctly, the kernel will truncate any data beyond 32k while the comment mentions that is the maximum length supported by a u16-type length field. I am in agreement that a u16 could support that long of a message but I do not think the kernel does based on the link I pasted.

On the topic of leaving allocation up to the user of library, I'm in 100% agreement the more I look into this. I am doing an API breaking change that will allow stack or heap allocation in whatever form the user desires as long as the struct they pass in implements the Read and Write traits. This will avoid eating up the stack with memory that will go unsused for shorter messages and that will probably cause concern in any kind of profiling of the code (32k is a lot to allocate on the stack after looking at stack limit defaults on Linux). This will also make it more flexible in the use case of allocating one 32k block on the heap and being able to pass it around as needed without reallocation.

faern:
@jbaublitz I don't understand your argument about `Read`/`Write` saving the user from allocating a lot if the message is short. Indeed if the user knows the message will be short they can allocate a small buffer on the stack. But `neli` will not be able to grow a stack buffer for the caller in case it does not fit. And even for heap allocation I doubt it will be beneficial performance wise to do a small allocation and then increase it if the message does not fit. Allocating again and copying the already read memory is quite expensive. And the caller probably has a fairly good idea how large the messages for their use case will be anyway.

Yes, it's probably not a good idea to have the buffer on the stack if it will be passed around a lot to different functions. But that is why it's left to the caller to decide. If they are not passing around the buffer to different functions they might want this possibility.

jbaublitz:
@faern I see where the confusion is. I misremembered the constraints on the Cursor type for `Read` and `Write`. My idea is to take a Cursor around allowing any struct that satisfies `AsRef<[u8]>` to be passed into this cursor. This gives flexibility both in terms of stack/heap allocation but also flexibility in representation of the data being passed into the Cursor for serialize and deserialize functions. I'm going to close this PR as I think this will take a larger rework of the code base to achieve more flexibility for the user of the library. Thank you for your contributions and I hope to have the next release ready soon.

jbaublitz:
@faern Just a heads up that I've pushed some breaking changes to the `develop` branch. Also I apologize if either of the previous two comments were confusing as they both were not accurate. I was skimming the documentation as I didn't quite have the time to focus on this as heavily.

The basic idea of the new trait definition is to give three options in an enum that replaces the less flexible `SerState` and `DeState`: a slice (stack buffer), a vector (growable heap buffer), or a boxed slice (sized heap buffer). Also I added in stateful and stateless serialization/deserialization methods (`serialize` vs. `serialize_with`) to replace the less clear internal state with `.set_usize`.

Thank you for the suggestions - this was a lot easier to work with using my own trait internally in the library and hopefully it will be for you too. Please feel free to open new issues or PRs and I should be testing out the develop branch and merging to master and releasing soon.

faern:
@jbaublitz I'm not sure I understand your distinction between the types of buffers. A slice is not strictly a stack buffer. A slice is just a pointer and length into a borrowed memory area:
```rust
fn main() {
    let vec_on_heap = vec![0; 1024];
    let array_on_stack = [0; 1024];

    // These two slices have identical types in the Rust type system.
    // But one points to memory on the heap and one on the stack.
    let slice_to_vec_on_heap: &[u8] = &vec_on_heap[..];
    let slice_to_array_on_stack: &[u8] = &array_on_stack[..];
}
```
So I don't think the `neli` abstraction for a buffer has to know anything about stack vs heap. It should be controllable by the caller, but does not need to show up in the types in `neli`. I think what `neli` has to know is if the buffer it has is growable (`Vec<u8>`) or not (`&[u8]`).

Also, what functionality does the boxed slice add? It's exactly the same as a normal slice, but it's owned. Or another way to look at it is that it's identical to a `Vec` but not resizable.

jbaublitz:
@faern I understand that it's not entirely accurate to say that slices always point to stack memory but that is the typical use case in which I see someone wanting a slice vs. a vector. Regardless, I discussed the boxed slice with one of my coworkers and this does seem unnecessary particularly given that if someone is looking for a sized heap reference, a `Vec`'s `as_mut_slice` or `as_slice` methods will accomplish this and can be passed into `MemRead`/`MemWrite`'s `new_slice` method. I'm removing this option in the next commit. Other than that, I think it's ready for the next release so that'll be coming soon.

jbaublitz:
@faern New release has been uploaded to crates.io (0.2.0). Thank you for your feedback/work with the `libc` crate.

PR #4, opened 2018-02-19
Make serialize work without mutating input

faern:
I noticed `self.push('\0');` in the serialization of `String`. I personally found it to be a strange property that serializing a type could possibly modify it. Because serializing the same instance multiple times could then yield different results. In the case of `String`, calling `Nl::size` on a `String` would differ before and after the serialization.

Here I make `&mut self` into `&self` in `Nl::serialize` to prohibit mutation of `self`. The only type affected by this was indeed `String`, which I solved by re-allocating the string buffer into one with the terminating null in the serialization implementation.

I was also looking into making the `deserialize` impl for `String` make use of `CString`. But I was not really sure how the lengths of the deserialization state worked. Especially this line:
```rust
let input = state.get_usize().unwrap_or(state.0.get_ref().len());
```

### DIFF
--- a/src/err.rs
+++ b/src/err.rs
@@ -40,6 +40,13 @@ impl Error for NlError {
 #[derive(Debug)]
 pub struct SerError(String);
 
+impl SerError {
+    /// Create a new error with the given message as description
+    pub fn new<T: ToString>(msg: T) -> Self {
+        SerError(msg.to_string())
+    }
+}
+
 try_err_compat!(SerError, io::Error);
 
 impl Display for SerError {
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -61,7 +61,7 @@ macro_rules! impl_var {
         }
 
         impl Nl for $name {
-            fn serialize(&mut self, state: &mut NlSerState) -> Result<(), SerError> {
+            fn serialize(&self, state: &mut NlSerState) -> Result<(), SerError> {
                 let mut v: $ty = self.clone().into();
                 try!(Nl::serialize(&mut v, state));
                 Ok(())
--- a/src/genlhdr.rs
+++ b/src/genlhdr.rs
@@ -48,7 +48,7 @@ impl Default for GenlHdr {
 }
 
 impl Nl for GenlHdr {
-    fn serialize(&mut self, state: &mut NlSerState) -> Result<(), SerError> {
+    fn serialize(&self, state: &mut NlSerState) -> Result<(), SerError> {
         self.cmd.serialize(state)?;
         self.version.serialize(state)?;
         self.reserved.serialize(state)?;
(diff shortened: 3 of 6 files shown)


src/err.rs
```diff
@@ -38,7 +38,7 @@ impl Error for NlError {
 
 /// Serialization error
 #[derive(Debug)]
-pub struct SerError(String);
+pub struct SerError(pub String);
```
faern: I had to make this public in order to create a custom error for the new type of string serialization error. There are probably other solutions. But I wanted something simple for now to present my idea here.

faern:
I'm looking at this part of string deserialization:
```rust
let pos = v.iter().position(|b| *b == 0);
if let Some(p) = pos {
    v.truncate(p);
}
```
So this means that if we read a set of bytes from the deserialization context, but it has a null byte in the middle, then we'll only produce a string from the bytes up until the null, but the rest are discarded, since they have already been pulled out of the `NlDeState`? As I said higher up I did not get exactly how the length of `NlDeState` worked, but would it not be an error if the length does not match the content? Feels like it can cascade in other problems down the road if we just throw away data.

jbaublitz:
First off, thank you for investing the time to look at these things. This was actually my first Rust project so it's great to get feedback on it. That being said, thank you for the catch here - there are definitely some dark corners that I need to revisit after having worked in the language for a longer period of time at this point. That `&mut  self` in serialize should definitely not be there and the behavior describing is defintely not what I want. As for the second comment, if Rust detects a null byte in a byte buffer that is not at the end of the `Vec` or slice, the conversion to a string fails. I have tried many ways around this but the only way to get around this seems to be to read all of the zeros and then remove all but the first one for compatibility with the C ABI. That or using unsafe code which may be the answer here? I'm not convinced that unsafe code is the best option though in this case.

faern:
@jbaublitz I see, welcome to the language/community! :)

So don't you want to fail if the buffer contains a zero byte? If the deserialization says "Here is a string: `[65, 0, 65, 0]`, go ahead and use it". Then whoever serialized that made an error, since that is not a correct C string. So instead of saying "Ok, here is the string "A"", maybe it should say "Nope, this is not a valid Netlink message".

jbaublitz:
So that's an issue with Netlink's format. It pads with zeros to meet the word length requirements and actually corrupts headers that it sends back if this requirement is not met. Let me think about if there's a better way to do this because as I remember, the netlink attribute length should only include the first null terminator in the size so I may be able to slice the string from the beginning of the payload until the byte identified by `nla_len` as the end. If I can find a way to use that instead, that should solve the problem without an iterator solution which always felt hacky. If you change the `pub String` into an ```
impl SerError {
    pub fn new<D>(msg: D) -> Self where D: Display {
        SerError(msg.to_string())
    }
}``` or something comparable so that the user can pass in both string literals and formatted String structs as an error message, I'm happy to merge this and then work on getting the iterator out of the equation.

faern:
Done. However, I have one concern with the `Nl::size` implementation for `String`. It does not include the null byte. I'm not entirely sure it should or not. There was no test for it. I added a test and "fixed" the implementation, is this correct?

jbaublitz:
You are correct, it should take into account the null byte. My guess for why this worked previously is because it was being called internally after the mutation which now that we've removed this, needs to be explicit in the `size` declaration. If you're ready, I will merge and start poking at the null terminator problem.

PR #5, opened 2018-02-20
Fix invalid NlFamily::Xfrm value

faern:
Fixing a copy-paste error introduced by me in #1.

I also realize another caveat with this enum. I introduced the last variant `InetDiag => libc::NETLINK_SOCK_DIAG` since it was present in the netlink header. The problem is that it has the same value as another variant, `SockDiag`. So when converting from enum to primitive it works fine. But converting it back will always yield the first variant. So `NlFamily::from(NlFamily::InetDiag.into())` will yield `nlFamily::SockDiag`, which might not be optimal.

I noticed this when experimenting with using https://crates.io/crates/derive-try-from-primitive to do the derivation, instead of crafting a homegrown macro. Then I got an error because Rust can't have multiple variants with the same value when you assign them directly to the variants.


faern:
These constants in `netlink.h` are not enums, so it's nothing wrong with having multiple constants with the same value. In this context they are basically addresses to different kernel parts, right? So both `InetDiag` and `SockDiag` have the same address. So modeling it with an enum is slightly wrong, which cause these inconsistencies with duplicates.

Since in C this would hold: `NETLINK_SOCK_DIAG == NETLINK_INET_DIAG`.
But in Rust this would not hold: `NlFamily::SockDiag == NlFamily::InetDiag` (with the current impl).

jbaublitz:
@faern Hm I'm not sure I agree that modelling with an enum is wrong. From a type perspective, you lose the ability to assure only a subset of constants are used in the proper context if you rip out the enums. My recommendation here is to create an enum that represents both of these constants explicitly in the name. From a determinism perspective, this will solve the issue. Both conversion from and to the variant will yeild consistent results.

Issue #11, opened 2018-12-31
Create an `impl Nl for u64` to help support `u64`s in Netlink responses


Dieff:
In `src/lib.rs` there are implementations of the `Nl` trait for a number of common types. An unsigned 64 bit integer (`u64`) is another common type that does not have an yet have implementation. This type can be found in several Netlink apis, including [Wireguard](https://github.com/WireGuard/WireGuard/blob/master/src/uapi/wireguard.h) (see the attribute `WGPEER_A_RX_BYTES`). It would help developers using neli to interact with these apis if this impl was provided by the library.

The underlying libraries used by neli seem to provide the methods you would need to create `impl Nl for u64` more or less the same way as the the `impl Nl for u32` and `impl Nl for u16` that already exist in neli. It should be a simple matter to add the support for `u64` as well.

I think I will need this to go forward with a project I am working on now: interfacing with the Wireguard kernel api (generic netlink) using neli. I could submit a pull request with my impl added to this library.

I would love some feedback about adding support (`Nl` `impl`s) for this type and other Wireguard related types to Neli.

jbaublitz:
@Dieff Hey there! I'm happy to accept a PR or work this into the 0.4.0 release. Either way, just let me know which you would prefer so we don't duplicate work.

jbaublitz:
If you would like it as soon as possible, let me know and I'll branch off of the last release and add a revision release and push that to crates.io then incorporate that into 0.4.0. Unfortunately I've been pulled in a few different directions so that release is taking a bit longer than I'd hoped.

Dieff:
I think I'll take a stab at writing the impl as it seems pretty simple and it would be nice for me to have quickly. I'll open a pr when I get it working.

jbaublitz:
#12 merged.

As discussed, please open another issue regarding the `impl Nl` discussion as I'd definitely like to resolve that with you. Thanks!

PR #12, opened 2018-12-31
Created `impl Nl for u64`

Dieff:
One of the common types found when dealing with Generic Netlink is `u64`. This PR adds an implementation of the `Nl` trait for `u64` so that these values can be easily deserialized by `neli` when reading Netlink messages.

You can use the new `impl` to parse Netlink attrs link this:
```
    let peer_attr_handle: neli::nlattr::AttrHandle<'_, u16> = ...;
    let recieved_bytes: u64 = peer_attr_handle.get_payload_with::<u64>(PEER_RECIEVED_BYES_ATTR, None).unwrap();
```

The implementation is defined in basically the same way as the other impls for unsigned integers are. 
A test that works like the others was also added to `src/lib.rs`.

This pr addresses #11.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -150,6 +150,24 @@ impl Nl for i32 {
     }
 }
 
+impl Nl for u64 {
+    type SerIn = ();
+    type DeIn = ();
+
+    fn serialize(&self, mem: &mut StreamWriteBuffer) -> Result<(), SerError> {
+        mem.write_u64::<NativeEndian>(*self)?;
+        Ok(())
+    }
+
+    fn deserialize<T>(mem: &mut StreamReadBuffer<T>) -> Result<Self, DeError> where T: AsRef<[u8]> {
+        Ok(mem.read_u64::<NativeEndian>()?)
+    }
+
+    fn size(&self) -> usize {
+        mem::size_of::<u64>()
+    }
+}
+
 impl<'a> Nl for &'a [u8] {
     type SerIn = ();
     type DeIn = &'a mut [u8];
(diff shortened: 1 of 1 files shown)


jbaublitz:
@Dieff This looks good. Before I merge this, I just want to make sure you're okay with what's been pushed to master recently. What's the most recent version of neli you're using? I've been using master for v0.4.0 so there are some backwards incompatible changes with 0.3.x. If this is a problem, We should organize around getting a release branch set up for a revision release and doing a PR targeting that.

jbaublitz:
This is an excellent reason for me to resurrect the dev branch and will do that after the `0.4.0` release so master stays frozen at the latest release and dev is future release work.

Dieff:
I've updated the dependencies in my project to use my fork of your latest master so that won't cause any problems for my project. I plan to upgrade to `0.4.0` once it's released on crates.io.

I'd be happy to open any future PRs against the 'dev' branch. IMO that makes sense as a way to do future work on `neli`.

Dieff:
Disclaimer: maybe this should be a separate issue. This PR does raise a question for me. There are a number of types that I have run into working with Netlink so far that I think could potentially have `impl Nl for ...`s such as `struct sockaddr_in` and `struct timespec`. When a user of `neli` gets these values as part of a Netlink message and wants to deserialize them how are they expected to do this? It seems to me that there are basically two options: create an implementation of `Nl` for that new type and then call `nl_attr::AttrHandle::get_payload_with` or access the raw payload via `nl_attr::AttrHandle::get_atribute().payload`. The first way makes the deserialization code much cleaner IMO.

When you create a new implementation for `Nl` and a new type outside of Neli and in your own code you run into Rust's [rules about orphan impls](https://internals.rust-lang.org/t/revisit-orphan-rules/7795). So if you wanted to define `impl Nl for libc::timespec` for example (I'm pretty sure this makes sense to do? Please correct me if it does not) then you have to create your own definition for `libc::timespec` instead of using the one from the `libc` crate. Maybe there is some way that this could be easier?

jbaublitz:
@Dieff First off, yes, I'd love to discuss the above problem in a separate issue. Please open one and we can discuss how we'd like to resolve it there. I can't imagine you're the only one who will bump into this so I'd like to come up with a story for how to actually address this and document this with your help. Thanks for bringing this to my attention!

Secondly, if you're okay with using master for the time being, I will do some quick testing on my end and then merge. Thank you for your contribution!

PR #14, opened 2019-01-14
Updated `consts` for better documentation

Dieff:
When using this library, I found the system for creating `enum`s that mapped to C constants very helpful when working with new Netlink apis. I had a hard time figuring out how to use it, though, because there wasn't much documentation of the macros, and there was no documentation with the individual constants in `neli::consts::*`. I tried my best to improve the docs so that the constants and macros would be easier to use.

Major Changes:

 * Updated the three constant generating macros with alternate versions that allow
doc comments as the first argument. You can use the macros the same way they were used before, or you can call them with the alternate version so that `cargo docs` shows the constants with the documentation that they are supposed to have.
* Moved the documentation about `impl_var!`, `impl_trait!`, and `impl_var_trait!` to actually live with the macros, instead of at the top of `consts.rs`. This fixes an error `error: missing documentation for macro` when building the library with `cargo 1.33.0-nightly (34320d212 2019-01-03)`. 
* Expanded the documentation for `impl_var!`.

I could use feedback on the new versions of the macros. They should be more convenient to use. Also the individual comments on members of `neli::consts::*` should probably be improved.

It would be good to have a way to specify comments on the individual fields in the constant enums as well.

PR #16, opened 2019-04-17
Update consts.rs for ability to autogenerate documentation for macro-defined netlink constants

jbaublitz:
Closes #14

### DIFF
--- a/src/consts.rs
+++ b/src/consts.rs
@@ -218,14 +218,14 @@ pub fn alignto(len: usize) -> usize {
 }
 
 impl_var!(
-    /// Address families 
+    /// Internet address families 
     ( Af, libc::c_uchar,
     Inet => libc::AF_INET as libc::c_uchar,
     Inet6 => libc::AF_INET6 as libc::c_uchar )
 );
 
 impl_var!(
-    /// Address families for sockets
+    /// General address families for sockets
     (  AddrFamily, libc::c_int,
     UnixOrLocal => libc::AF_UNIX,
     Inet => libc::AF_INET,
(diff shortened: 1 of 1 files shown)


jbaublitz:
Hi @Dieff, let me know if you have any additional input before I merge. Thanks so much for putting the time into doing that!

Dieff:
It's been a while since I have worked with this library, I'm hoping to do so again now though. From the perspective of a developer using it as a dependency, having additional information on the docs.rs page is a big help.

Are you happy with the new syntax for the macros? I felt it was a bit awkward having to put the comment inside the macro, instead of putting it above the definition like with most doc comments. With the current approach you write:
```
impl_var!(
  /// test doc comment
  (MyConst, u16,
    Id => 1,
    Name => 2,
  )
);
```
But I think it would be more convenient to be able to write
```
/// test doc comment
impl_var!(MyConst, u16,
    Id => 1,
    Name => 2,
);
```

 I am pretty inexperienced with the Rust macro system and couldn't work out a way to include a comment from outside the macro in the generated code (if this is even possible?), so I went with the former method.

If the way the macros work here is good enough, I don't have anything to add at the moment. I hope to use the new system to PR some new `libc` values and constants for Wireguard related Netlink the near future.

jbaublitz:
I can say that one way around this is procedural macros. I've written them before and can make that happen if that's what you prefer? It could make this cleaner for sure but I'm also not sure it's worth the investment given that this works. It would provide more flexibility on annotating enum variants, that is definitely one benefit. I'll take a look at the macro you've provided in more depth and see if I can make it a little bit cleaner and more flexible and then report back here. Thanks again!

jbaublitz:
I made the syntax a little cleaner and removed the unnecessary parentheses that make the syntax a bit clunkier because of the macro system in Rust. This makes it look more like the enum it generates and is probably less likely to cause headaches for beginners using this macro. Going to merge.

Issue #15, opened 2019-04-13
Handle multiple netlink messages in single response


gluxon:
I'm not sure what circumstances this occurs under, but occasionally the kernel can return multiple Netlink messages in the same `libc::recv` call. When this happens, neli interprets the second packet's bytes as an attribute of the first message (which is incorrect).

https://github.com/jbaublitz/neli/blob/1925ba1b3d876f981d0817bbf28ec22b311e6dcf/src/socket.rs#L140-L148

IMHO, `NlSocket::recv` should interpret the first few bytes of the packet, get the length of the next message, and only read that length until the next call to `NlSocket::recv`.

jbaublitz:
Hello! Thanks for opening an issue. I'm just starting back up with this project after not having been able to look at it for a bit.

A few notes:
1. Netlink can indeed have a netlink headers nested within an attribute of another message as is the case with error messages. An error message or an ACK at the protocol level will return a message with a certain `nl_type` indicating whether it is an ACK or error message and will then return the header of the request that generated the error or ACK response. This may appear to be two separate messages but actually is not and the request header is nested under an attribute in the response. Unless you can provide specific scenarios where this is reproduceable, I'm going to assume you're referring to the above scenarios which is defined in the netlink docs.
2. If I am incorrect, the suggestion to only read the first message and then the second is unfortunately not possible with netlink sockets as any data sent in the same "chunk" must be read into a buffer in the same read call or the remaining data that is not read is lost. As I understand it, this is the same as with UDP. As a result, the method of handling netlink messages that I default to is to overestimate the amount of bytes that will come back as the `recv` syscall will handle less bytes being available than the size of the specified buffer but will lose data if underestimated. I've tested this on a number of occasions and this behavior is consistent when reading from a file descriptor into a user space buffer.
3. If point 1 is incorrect, my suggestion for remedying the situation would be, after reading the both messages into a buffer, to simply check if the size of the total parsed netlink message is less than the number of bytes read.

In summary, I do think there would be a parsing error based on the way I handle parsing attributes from the binary representation if this were not indeed one of the situations in point 1 but am open to reproducing and fixing if you can provide me with an example of a scenario where this is not the case. 

Thanks!

gluxon:
> Hello! Thanks for opening an issue. I'm just starting back up with this project after not having been able to look at it for a bit.

Thanks for the response! Didn't expect one so quickly. I appreciate the work you've done on this library and it's definitely been super useful to me.

[...]

PR #18, opened 2019-05-11
Add genl ctrl list example

gluxon:
Followup from: https://github.com/jbaublitz/neli/issues/15#issuecomment-490627034

### DIFF
--- a/examples/ctrl-list.rs
+++ b/examples/ctrl-list.rs
@@ -29,8 +29,8 @@ fn main() -> Result<(), NlError> {
     };
     socket.send_nl(nlhdr)?;
 
-    loop {
-        let response = socket.recv_nl::<Nlmsg, Genlmsghdr<CtrlCmd>>(None)?;
+    let mut iter = socket.iter::<Nlmsg, Genlmsghdr<CtrlCmd>>();
+    while let Some(Ok(response)) = iter.next() {
         match response.nl_type {
             // This example could be improved by reinterpreting the payload as an Nlmsgerr struct
             // and printing the specific error encountered.
--- a/examples/ctrl-list.rs
+++ b/examples/ctrl-list.rs
@@ -30,17 +30,14 @@ fn main() -> Result<(), NlError> {
     socket.send_nl(nlhdr)?;
 
     loop {
-        let response = socket.recv_nl::<GenlId, Genlmsghdr<CtrlCmd>>(None)?;
-
-        if let GenlId::UnrecognizedVariant(id) = response.nl_type {
-            match Nlmsg::from(id) {
-                // This example could be improved by reinterpreting the payload as an Nlmsgerr
-                // struct and printing the specific error encountered.
-                Nlmsg::Error => panic!("An error occurred while retrieving available families."),
-                Nlmsg::Done => break,
-                _ => {}
-            }
-        }
+        let response = socket.recv_nl::<Nlmsg, Genlmsghdr<CtrlCmd>>(None)?;
+        match response.nl_type {
+            // This example could be improved by reinterpreting the payload as an Nlmsgerr struct
+            // and printing the specific error encountered.
+            Nlmsg::Error => panic!("An error occurred while retrieving available families."),
+            Nlmsg::Done => break,
+            _ => (),
+        };
 
         let mut handle = response.nl_payload.get_attr_handle::<CtrlAttr>();
         handle.parse_nested_attributes()?;
--- a/examples/ctrl-list.rs
+++ b/examples/ctrl-list.rs
@@ -1,0 +1,67 @@
+extern crate neli;
+use neli::consts::{CtrlAttr, CtrlCmd, GenlId, NlFamily, NlmF, Nlmsg};
+use neli::err::NlError;
+use neli::genl::Genlmsghdr;
+use neli::nl::Nlmsghdr;
+use neli::nlattr::Nlattr;
+use neli::socket::NlSocket;
+use neli::Nl;
+use neli::StreamReadBuffer;
+
+const GENL_VERSION: u8 = 2;
+
+// This example attempts to mimic the "genl ctrl list" command. For simplicity, it only outputs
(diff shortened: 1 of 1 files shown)


examples/ctrl-list.rs
```diff
@@ -0,0 +1,67 @@
...
+    };
+    socket.send_nl(nlhdr)?;
+
+    loop {
+        let response = socket.recv_nl::<GenlId, Genlmsghdr<CtrlCmd>>(None)?;
+
+        if let GenlId::UnrecognizedVariant(id) = response.nl_type {
+            match Nlmsg::from(id) {
+                // This example could be improved by reinterpreting the payload as an Nlmsgerr
+                // struct and printing the specific error encountered.
```
gluxon: I couldn't figure out an idiomatic way to do this with the library at the moment. If I missed something, let me know if there's a way to do this within neli and I can add the behavior mentioned in this comment.

jbaublitz: Ah! I see the misunderstanding. If you do the following, this might be easier.
```
let response = socket.recv_nl::<Nlmsg, Genlmsghdr<CtrlCmd>>(None)?;
match response.nl_type {
    Nlmsg::Error => {
        println!("An error occurred while retrieving available families");
        std::process::exit(1);
    },
    Nlmsg::Done => break,
    _ => (),
};
```
Because `Nlmsghdr` accepts a type parameter that can be satisfied with any type `T` as long as `T` is marked as compatible using `NlType`. `Nlmsg` is marked as `NlType`-compatible [here](https://github.com/jbaublitz/neli/blob/master/src/consts.rs#L497) so it can be passed as the first type parameter to `recv_nl` calls and return an `Nlmsghdr` with `nl_type` deserialized as the type `Nlmsg`. Does that make sense? I think this could be documented better at a high level so I'll make a note to do that.

gluxon: That makes sense. I'll make that change since it definitely simplifies this chunk of code.

The comment I left concerns the second type parameter more though. I was hoping to print out the error code within [`nlmsgerr`](https://github.com/jbaublitz/neli/blob/56bd6c7a180c780a4c7e13c3a70f9f9801237545/src/err.rs#L28), but that seems difficult since the buffer was already consumed as a `Genlmsghr`.

jbaublitz: After reviewing this further I think I see why you're doing it the way you're doing it - is it because you aren't sure whether a response is going to come back as `Nlmsg` type or `GenlId`? If that's the case, because you aren't actually using the `nl_type` field anywhere except to check for `Nlmsg::Done`, it may be worth doing what I suggest above. Typically what I do if I'm not sure what type is coming back and need to handle both scenarios is simply pass in `u16` as the type parameter (I've added the ability to use a raw `u16` as the type) and then try parsing it as both in whatever makes sense logically after it's been deserialized. It's less elegant but sometimes it makes more sense due to netlink being a heavily integer and C-targeted protocol. I'd like to make this less of a manual process but still am not sure how to make that better than what I just suggested at the moment.

gluxon: > After reviewing this further I think I see why you're doing it the way you're doing it - is it because you aren't sure whether a response is going to come back as Nlmsg type or GenlId?

Yes, that's exactly it. But not only are the types undetermined until response time, the payload is as well.

> If that's the case, because you aren't actually using the nl_type field anywhere except to check for Nlmsg::Done, it may be worth doing what I suggest above. Typically what I do if I'm not sure what type is coming back and need to handle both scenarios is simply pass in u16 as the type parameter (I've added the ability to use a raw u16 as the type) and then try parsing it as both in whatever makes sense logically after it's been deserialized.

Were you referring to the type or the payload here? I'm not aware of any way to parse a buffer as both for the later case. I agree it's trivial for the former (type) case.

> I'd like to make this less of a manual process but still am not sure how to make that better than what I just suggested at the moment.

Hm.. I think a Rust idiomatic way would be to define an enum that extends Nl.

```rust
#[derive(Nl)]
enum PossibleResponse {
  CtrlMsg(Genlmsg<CtrlCmd>),
  Nlmsgerr(Nlmsgerr<CtrlCmd>),
}
```

and then

```rust
let response = socket.recv_nl::<Nlmsg, PossibleResponse>(None)?;
```

I'll play with this method a bit.

gluxon:
The new syntax with `.iter()` looks lovely. I'll definitely incorporate that into this example.

examples/ctrl-list.rs
gluxon: > Hm.. I think a Rust idiomatic way would be to define an enum that extends Nl.

Played around with it a bit but didn't get very far. We need a way to map message type ids to deserialization methods to do this, which is probably not within the scope of this pull request.

gluxon:
So I understand why you brought up the new `.iter()` method. Was there something pertaining to the changes to `.recv_nl()` and `.recv_ack()` that could be used to improve the pull request?

jbaublitz:
In this case, I was just letting you know about the API as a whole for future reference but I do agree that `.iter()` is most likely best here. I see the use case for `.recv_nl()` and `.reck_ack()` as more being a lower level approach where the types of the messages being returned aren't always the same. `.iter()` is helpful for streams of notification or event-based message streams where the type is the same. `.recv_nl()`/`.recv_ack()` seems to be more helpful for request/response workflows using the same socket repeatedly.

examples/ctrl-list.rs
jbaublitz: I agree that this is outside the scope of this pull request but I made an issue #19 so we can discuss more there. This could be really beneficial for less manual work for users in parsing raw integers so I'm happy to move the discussion over there. I have some ideas for this and would love to bounce them off you before implementing. Thanks!

jbaublitz:
I'm going to merge this and make a few changes post merge. Thanks so much!

gluxon:
> I'm going to merge this and make a few changes post merge. Thanks so much!

Thank you!

PR #24, opened 2019-07-07
Use get_nested_attributes R type for message type

gluxon:
The template type for `get_nested_attributes` was previously used in neli to specify the Netlink attribute type for the nested attributes. It's currently unused in neli 0.4.2. I think this may have been a typo since forcing all nested types to be T makes parsing nested attributes of a different type than the parent impossible.

If I'm wrong, please feel free to close the pull request.


jbaublitz:
I found a similar error elsewhere that should be corrected now so thank you for tracking this one down too. I'm going to merge this.

PR #25, opened 2019-07-08
Update nla_len when adding a nested attribute

gluxon:


jbaublitz:
Hey, good catch here, thank you! I have two requests for this PR. Would you be willing write a test for this so it does happen again if code changes? Also I'm playing with the idea of changing both [this line](https://github.com/jbaublitz/neli/pull/25/commits/ad108fcabb8d1640aa2478ffc7dd8de1e4ee641c#diff-74c894e861acc2994fb111c84a47901bR113) and [this line](https://github.com/jbaublitz/neli/pull/25/commits/ad108fcabb8d1640aa2478ffc7dd8de1e4ee641c#diff-74c894e861acc2994fb111c84a47901bR121) to something like:
```rust
// Put this after:
// let size = self.payload.len() as u64;
// let mut buffer = StreamWriteBuffer::new_growable_ref(&mut self.payload);
// buffer.set_position(size);
// attr.serialize(&mut buffer)?;
self.nla_len = self.size() as u16;
```
I'm finding that I have a lot of length errors popping up throughout the code and I want to standardize on a way of calculating length so that if something changes, I don't have to track down all of the places I need to change behavior. If `.size()` is implemented correctly, it should calculate the length of the payload with no padding. Does that sound good to you?

gluxon:
> Would you be willing write a test for this so it does happen again if code changes? 

Yes, good idea!

> If .size() is implemented correctly, it should calculate the length of the payload with no padding. Does that sound good to you?

Sounds good, but why isn't `size()` for `Nlattr<Vec<u8>>` simply `self.nla_len`? At the moment, the return for `size()` includes the payload's length, which has padding added to it.

https://github.com/jbaublitz/neli/blob/1670a1659b1b920c3ea12596027e18e22be8c805/src/nlattr.rs#L173-L175

src/nlattr.rs
```diff
@@ -311,4 +312,28 @@ mod test {
...
+
+        let mut nlattr = Nlattr::new::<Vec<u8>>(None, CtrlAttr::Unspec, vec![]).unwrap();
+        assert_eq!(nlattr.size(), 4);
+
+        nlattr.add_nested_attribute(create_aligned_attr()).unwrap();
+        assert_eq!(nlattr.size(), 12);
+
+        nlattr.add_nested_attribute(create_unaligned_attr()).unwrap();
+        // Commented out because the following fails at the moment.
+        // assert_eq!(nlattr.size(), 17);
```
gluxon: So I expect this to be `17`. Do you John?

I don't think the Netlink specification outlines what the length field of an outer attribute should be when it contains inner attributes that are padded. Ex: If an inner attribute has length `5`, I expect an outer attribute (with only that singular inner attribute) to have length `9`. But maybe the outer attribute having `12` is also an okay answer.

jbaublitz:
> Sounds good, but why isn't size() for Nlattr<Vec<u8>> simply self.nla_len? At the moment, the return for size() includes the payload's length, which has padding added to it.

As it stands, this is not accurate because of how I reworked padding. Based on some of the bugs that popped up in my handling of padding that you pointed out along the way, I've reworked it so attribute payloads are padded but individual data types are not. This means that you can always rely on `neli` to strip padding for `Vec`s and such when deserializing and padding will not be included when serializing until it is encapsulated in an `Nlattr` struct. This allows usage of the `size()` function in all cases for the length without padding.

Now you do bring up an interesting point about whether there should be padding at the end of the last attribute. I agree that this may seem unnecessary but does not break the protocol when communicating with the kernel. A few of my considerations when doing this were:
* It does work despite the potentially confusing way this is done
* It allows constant time additions of new nested attributes one-by-one
* It would not be constant time addition if the last attribute were not padded as the entire buffer would have to be re-parsed from the beginning to determine where the last attribute begins, what that last attribute's length would be without padding, and what the length of padding would need to be each time a new one is appended after the last attribute

As it stands, it works because each `Nlattr` is padded when serialized which guarantees the following:
* Attributes containing nested attributes will always be properly aligned because each attribute making up the payload is 4-byte aligned and the header is 4 bytes
* Data types in arbitrarily deeply nested attributes will be deserialized with padding stripped and serialized with padding added

src/nlattr.rs
jbaublitz: A better thing to test might be this. Make sure the outer attribute `size()` is aligned and the inner attribute `size()` is not aligned as long as it contains no nested attributes but only a serialized Rust primitive. Does that make sense? I posted a longer explanation in the main thread on this PR.

gluxon: Kind of. Are you saying a new/different test that contains the following?

```rust
let nlattr = Nlattr::new(None, CtrlAttr::Unspec, vec![1]).unwrap();
assert_eq!(nlattr.size(), 5);
assert_eq!(nlattr.asize(), 8);
```

gluxon:
Thanks for explaining John. I think we're in enough of an agreement to move on, but I do still have some questions for the sake of curiosity:

> This allows usage of the size() function in all cases for the length without padding.

It's this definition that confuses me. Isn't this the exact same invariant for `nla_len`? If so, why doesn't the body of `size()` always return `self.nla_len`?

> It would not be constant time addition if the last attribute were not padded as the entire buffer would have to be re-parsed from the beginning to determine where the last attribute begins, what that last attribute's length would be without padding, and what the length of padding would need to be each time a new one is appended after the last attribute.

I may be missing something. Suppose `nla_len`/`size()` returns the length of the nested attribute without the last attribute's padding bytes included (which isn't the case now). Then can't the serialization library simply add `asize() - size()` zeroes before serializing the next addition to that buffer?

jbaublitz:
> It's this definition that confuses me. Isn't this the exact same invariant for nla_len? If so, why doesn't the body of size() always return self.nla_len?

This is in part due to how I was calculating length previously and this is probably a good time to switch to what you're suggesting.

> I may be missing something. Suppose nla_len/size() returns the length of the nested attribute without the last attribute's padding bytes included (which isn't the case now). Then can't the serialization library simply add asize() - size() zeroes before serializing the next addition to that buffer?

Let me take more of a look to verify that this doesn't change anything in the way I handle attributes, but if you can get this working in this PR without breaking tests, I would definitely accept that change as I think that would probably be less confusing to contributors.

jbaublitz:
Hi Brandon, thanks for the test case! I've been using it to try a few things and it's been very helpful. After digging into this a bit more, I have a few thoughts:
1. Size should be defined in terms of the `.size()` of each of its components. The reason for this is that if it is not, tests begin failing and require some changes to fix. More generally, I'd prefer this approach as it allows us to trust `.size()` and `.asize()` to do the right thing even if someone sets `nla_len` explicitly and calculates the wrong value. It ultimately seems to be a "safer" alternative to using `.nla_len` directly which can be changed arbitrarily.
2. I checked into how we could potentially add padding at the end of the previous attribute when adding a new attribute but that would require a change in `Nlattr`'s serialization function. I'm not sure this adds a whole lot of value and does increase code complexity to some extent. If the current way of serializing were not handled by the kernel, I would absolutely agree that we should change this but as it does work, I think this might be the way it stays for now.

src/nlattr.rs
jbaublitz: More what I was thinking is this:
```rust
let mut nlattr = Nlattr::new(None, CtrlAttr::Unspec, vec![]).unwrap();
let subattr = Nlattr::new(None, CtrlAttr::Unspec, 5u16).unwrap();
nlattr.add_nested_attribute(subattr).unwrap();
assert_eq!(subattr.size(), 6);
assert_eq!(nlattr.size(), 10);
```

I would expect it to work this way because when attributes are not being nested, the length is provided without padding. As soon as we start nesting them, each attribute is added with padding leaving the `.size()`/`.asize()` for an attribute with nested attributes 4 byte aligned.

jbaublitz:
I now see what you were referring to about `.size()` including the padding earlier. I apologize for the misunderstanding. I've taken the time to really dig though padding and I think I need to come up with a better way to do this. The answer may be as simple as padding every struct representing a header/payload when serializing. I'm going to think some more on this. As always, I appreciate your help and your eyes on this.

jbaublitz:
Hi Brandon, I've built on top of your PR changes. Take a look at the branch [`gluxon-pr-rc`](https://github.com/jbaublitz/neli/tree/gluxon-pr-rc) if you're curious. I'm not sure if this suits your needs but I will continue working on it and making changes in that direction and adding tests, ultimately merging my branch incorporating your commit. Thanks for pointing this out. I'm really trying to clean up padding, etc and it is incredibly hard to do right in all cases in netlink. Changes are noted in [`CHANGELOG.md`](CHANGELOG.md) and I have ultimately decided to go with your approach of not appending padding when serializing an `Nlattr`. The pattern I'm seeing emerge is that anything represented by a `Vec`should pad in between elements but not after the final element, and once it is serialized all the way down, the top-level header should pad at the end to avoid the confusion you mentioned earlier in the PR. Does that all sound good?

jbaublitz:
This should be fixed in the next release (`v0.4.3-r1`). Please comment here if you have any other questions. `nlattr.rs` documentation is updated to explain my solution. Thanks for your contribution!

gluxon:
Good changes John.

I ended up searching `libnl` source code a bit more and it follows what you're doing. So you were right and my assumption of how netlink works with nested attribute lengths was wrong. `assert_eq!(attr.size(), 20)` was the right thing to do for this test case.

https://www.infradead.org/~tgr/libnl/doc/api/attr_8c_source.html#l00810
https://www.infradead.org/~tgr/libnl/doc/api/msg_8c_source.html#l00422

gluxon:
Thanks for looking into this! The documentation you added for this makes a lot of sense. 👍

Issue #19, opened 2019-05-13
Better way for parsing trait-marked constants in struct fields


jbaublitz:
The current way I get around the many constants that can be passed into certain fields like `Nlmsghdr.nl_type` is separating each set of constants into its own finite enum and marking the enum as compatible with this field by implementing the `NlType` trait for that enum. This is a good first step but a use case that came up in [this pull request](https://github.com/jbaublitz/neli/pull/18) made it clear that currently the best way to differentiate between different enums when parsing a stream where for example `nl_type` could be one of multiple enums is to use a `u16` as the type there and then manually try parsing as both enum types. This is not an ideal way to do this and I'm interested in figuring out a better way to do this.

jbaublitz:
I've been considering this and it may make the most sense to implement methods in the marker trait. Do you have any thoughts on this @gluxon?

gluxon:
> I've been considering this and it may make the most sense to implement methods in the marker trait.

Were you thinking of adding a deserialization method, or something else?

jbaublitz:
What I'm considering is some sort of approach that takes advantage of the fact that every enum that implements a given marker trait will have variants such that when each variant is converted to an integer, it will have a unique value across all variants in the set of variants for enums implementing the marker trait. This is guaranteed because each integer value must be unique in fields that are not bitwise flags which will never have marker traits. How to implement this is a bit trickier as doing it automatically sounds a bit more complicated but it may be possible to do manually with some `TryInto`-type code implemented for each marker trait. Does that sound reasonable?

gluxon:
So I was thinking of something like this, but it doesn't work since type ids are defined at runtime.

```rust
#[derive(neli)]
enum GenlResponse {
  #[type(3,4)] Nlmsg(Nlmsg),
  #[type(16)] CtrlCmd(CtrlCmd),
}
```

> How to implement this is a bit trickier as doing it automatically sounds a bit more complicated but it may be possible to do manually with some TryInto-type code implemented for each marker trait.

Yeah, I'm leaning towards this as well. Maybe a generic `NlPayload` type that has a vector of attributes. Other payloads then implement a `TryFrom` for this type.

jbaublitz:
Hi Brandon, I've been thinking on this for a bit and I'm pretty sure I have a solution. It's going to take some reworking of the macros but I'd like to do this now rather than down the road.

### Proposed solution
I'd like to remove the macro `impl_var_trait` and rework `impl_var` and `impl_trait` as follows.

[...]

PR #28, opened 2019-07-17
Small fixes for rtnetlink, and add route-list example to exercise it.

skrap:
The Rtattr types needed some help with their padding and size hints.  I tried to follow the responsibility rules you laid out in nlattrs.rs.  I added some tests as well.

I added a sample below, `route-list`, which is a cheap imitation of the `ip route` command.  Seems to work for me, but I've not tested it extensively!

### DIFF
--- a/examples/route-list.rs
+++ b/examples/route-list.rs
@@ -1,0 +1,116 @@
+extern crate buffering;
+extern crate neli;
+
+use buffering::StreamReadBuffer;
+use neli::consts::*;
+use neli::nl::NlEmpty;
+use neli::nl::Nlmsghdr;
+use neli::rtnl::*;
+use neli::socket::*;
+use neli::Nl;
+use neli::MAX_NL_LENGTH;
+use std::net::IpAddr;
+
+
+///
+/// This sample is a simple imitation of the `ip route` command, to demonstrate interaction with the rtnetlink subsystem.  
+/// 
+fn main() {
+    let mut socket = NlSocket::connect(NlFamily::Route, None, None, true).unwrap();
+
+    let rtmsg = Rtgenmsg {
+        rtgen_family: libc::AF_INET as u8, // Only ask about Ipv4, as it's simpler to display below.
+    };
+    let nlhdr = {
+        let len = None;
+        let nl_type = Rtm::Getroute;
+        let flags = vec![NlmF::Request, NlmF::Root, NlmF::Match];
+        let seq = None;
+        let pid = None;
+        let payload = rtmsg;
+        Nlmsghdr::new(len, nl_type, flags, seq, pid, payload)
+    };
+    socket.send_nl(nlhdr).unwrap();
+
+    'socket_read: loop {
+        let mut mem = vec![0u8; MAX_NL_LENGTH];
+        match socket.recv(&mut mem, 0) {
+            Err(e) => {
+                println!("Error: could not read from netlink socket: {}", e);
+                break;
+            }
+            Ok(x) if x <= 0 => {
+                println!("Error: could not read from netlink socket: {}", x);
+                break;
+            }
+            Ok(read_len) => mem.truncate(read_len as usize),
+        }
+        let mut buf = StreamReadBuffer::new(&mem[..]);
+
+        while !buf.at_end() {
+            // Provisionally deserialize as a Nlmsg first.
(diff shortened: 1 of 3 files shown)


jbaublitz:
Hi @skrap I don't see `Rtgenmsg` documented anywhere in `man 7 rtnetlink`. Is this a data structure you're creating to help with the process of talking to route netlink or one that you've found somewhere else? If this is simply a wrapper for one use case of passing a `u8` as a payload to `Nlmsghdr` as it appears it is, this is already supported without `Rtgenmsg` as `u8` implements `Nl` and any `u8` can be passed as a payload to `Nlmsghdr` for route netlink without this addition. Could you explain a little more why you've added this in case I'm missing something?

skrap:
Hi @jbaublitz !
Thanks for checking this out.

Re: Rtgenmsg:
It's in rtnetlink.h.  It's actually a prefix of most (all?) rtnetlink messages, and I'm guessing that it's perhaps vestigial.  The kernel seems to handle it, though.  I was following some oldish blog posts on how to query rtnetlink from C, and that's what they were using, but from `rtnetlink(7)` it seems like I could probably just use a `Rtmsg` instead, with all the fields filled with their wildcard-equivalent variants/values.  As for just using a `u8`, I thought about it, but since I was following an example which used rtgenmsg, I created the struct.  I'm not attached to it, and would happily remove it if you would rather not have it!

jbaublitz:
Thanks for the clarification! I'll dig around some more and see what makes the most sense here. If this is not deprecated, this would be great to include.

LuoZijun:
image

skrap:
I added one more bugfix, and successfully changed route-list to use the full-size Rtmsg rather than Rtgenmsg.

jbaublitz:
Hi @skrap - this looks great. Thank you for keeping `Rtgenmsg` as I would like to at least support that seeing as it's in the official kernel headers but I'm not sure from the documentation how exactly it's used (official netlink documentation is pretty hard to find for every use case it seems). However I do agree with your change. Based on what I know of generic netlink at this point, I do think `Rtmsg` needs to be the header after `Nlmsghdr`. For generic netlink, it's always `Genlmsghdr` unless there's an error from what I've seen. I'll either provide more detailed feedback or merge soon. Thanks for your contribution! Help with rtnetlink is much appreciated as it's the area I'm least familiar with in netlink.

examples/route-list.rs
```diff
@@ -0,0 +1,125 @@
...
+        let seq = None;
+        let pid = None;
+        let payload = rtmsg;
+        Nlmsghdr::new(len, nl_type, flags, seq, pid, payload)
+    };
+    socket.send_nl(nlhdr).unwrap();
+
+    'socket_read: loop {
+        let mut mem = vec![0u8; MAX_NL_LENGTH];
+        match socket.recv(&mut mem, 0) {
```
jbaublitz: So I'm reviewing now. Was there a reason that you used `recv` instead of `recv_nl` here? Much of the following lines look like they are a reimplementation of `recv_nl`.

src/rtnl.rs
```diff
@@ -31,8 +31,10 @@ impl<T, P> Nl for Vec<Rtattr<T, P>> where T: RtaType, P: Nl {
         )?;
         let mut vec = Vec::new();
         while size_hint > 0 {
-            let attr = Rtattr::deserialize(buf)?;
-            size_hint -= attr.asize();
+            let attr : Rtattr<T,P> = Rtattr::deserialize(buf)?;
+            size_hint = size_hint.checked_sub(attr.asize()).ok_or_else(|| {
```
jbaublitz: `.checked_sub()` is a good one to know about. There are some other places I should use this in `neli`.

jbaublitz:
I've pulled in your commits and committed on top of them in [branch v0.4.3-r2](https://github.com/jbaublitz/neli/tree/v0.4.3-r2). Thanks for your contribution!

examples/route-list.rs
skrap: It was to allow a reinterpretation of the received buffer, rather than discarding it, when looking at a message.

jbaublitz: Okay, that makes sense. I would recommend doing it the way I have done it until #19 is complete. That should ease some of this.

PR #39, opened 2019-10-01
Derive implementations

oherrala:
This PR contains two things:

 * Derive (Partial)Eq, (Partial)Ord and Hash traits for enums. Deriving these five traits allow using these enums in HashMaps and BTreeMaps.

* Derive Debug for structs in routing netlink subsystem. These make it much easier to figure out what happens by logging or print debugging.

### DIFF
--- a/src/consts/macros.rs
+++ b/src/consts/macros.rs
@@ -99,7 +99,7 @@ macro_rules! impl_var {
         $name:ident, $ty:ty, $( $( #[cfg($meta:meta)] )* $var:ident => $val:expr ),*
     ) => ( // with comments
         $(#[$outer])*
-        #[derive(Clone,Debug,Eq,PartialEq)]
+        #[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
         pub enum $name {
             $(
                 $(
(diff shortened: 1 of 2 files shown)


jbaublitz:
Hi @oherrala, thanks for the contributions! I'm going to have to take a look into trait bounds on struct type parameters again before merging this. I had a fairly active core Rust team member caution me about trait bounds on struct type parameters as opposed to the `impl`s on those structs in a specific scenario so I've avoided them elsewhere in `neli`. I'm going to do some digging, but as I recall when I did this, it introduced some compilation errors in some of the other modules. It looks like it passes CI here, but I want to dig back into what happened and why he recommended against this last time so that I don't cause myself problems down the line. I've bumped into places where they're necessary in the past, and I'd like to see if it makes sense to impose those restrictions elsewhere in `neli` or if I should keep it as is.

jbaublitz:
Please see this issue for a discussion of trait bounds of struct type parameters: https://github.com/rust-lang/rust-clippy/issues/1689. For now, I'd prefer to have those removed and I've opened an issue to evaluate whether this should indeed be the case in v0.5.0 (#43). I think I will probably maintain this requirement even in v0.5.0 simply because of the example of how easy it is to introduce breaking changes when defining trait bounds directly on struct type parameters. Once they're removed here, I'll merge this.

oherrala:
@jbaublitz Thanks! I force pushed new changes without the trait bounds. There was a reason I needed to add the trait bounds on structs, but now removing those I can't reproduce the issue. Dunno what the reason was, but now locally tests pass without!

jbaublitz:
Looks good to me. Thanks for contributing! I'm merging this now.

oherrala:
Thanks!

Issue #31, opened 2019-07-21
Allow Nlmsgerr to be returned as an Err type in a Result


jbaublitz:
@gluxon You may be interested in this. I'm planning to start checking for `Nlmsgerr` in netlink return packets and passing that back as an error. Two ways I can think of doing this - either implementing `Error` for this data type or adding a variant to `NlError` which wraps an `Nlmsgerr` when that is returned. Thoughts? This may make the parsing issue in #19 just about types instead of types and error handling as I feel this is outside the scope of that issue.

gluxon:
This definitely seems like an idea that would work, but I'm curious if it does completely eliminate any reasons to handle different response payload types (like we were talking about in #19).

I couldn't think of any problems in the last 2 weeks, so handling errors as a special case seems like a good way to go.

jbaublitz:
I've decided to add an `Error` implementation in `v0.5.0-rc1` but it won't be take advantage of until the major breaking changes in `v0.5.0-dev`. My decision to use the library `buffering` has put me in a position where I can't do very much to make this usable in `v0.5.0-rc1` as it requires so parsing of individual fields of messages in a way that just isn't possible.

jbaublitz:
Mostly resolved by bac0f02c1b1ed6d4f0dff7f6f1cb835964f74422. I'll leave this open until I add the resolution to branch `v0.5.0-dev`.

jbaublitz:
This is now resolved in `v0.5.0-dev`.

PR #42, opened 2019-10-07
Nlmsgerr can implement debug

vorner:
It comes handy when playing with the protocol.

It would be even better if it could also implement the `Error` and `Display` traits. It seems the error is negative errno, so it could be resolved by using the `std::io::Error`. Is it OK to change the struct to contain that instead of the direct error code? Or maybe add an `Into` implementation?

### DIFF
--- a/src/err.rs
+++ b/src/err.rs
@@ -41,6 +41,7 @@ macro_rules! try_err_compat {
 }
 
 /// Struct representing netlink packets containing errors
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


jbaublitz:
There's already an open issue for implementing `Error` for `Nlmsgerr` (#31). This should be done by 0.5.0. As for adding `Debug` in the meantime, this makes sense. I'll merge this.

PR #44, opened 2019-10-08
Various fixes

vorner:
Collection of little things that I, however, believe to be wrong. Should I go into more details why I think so?

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -344,7 +344,7 @@ impl NlSocket {
         if let Some(seq) = self.seq.as_mut() {
             *seq += 1;
         }
-        if let Some(true) = self.buffer.as_ref().map(|b| b.at_end()) {
+        if self.buffer.as_ref().map(|b| b.at_end()).unwrap_or(false) {
             self.buffer = None;
         }
         Ok(msg)
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -341,8 +341,8 @@ impl NlSocket {
         } else if self.pid != Some(msg.nl_pid) {
             return Err(NlError::BadPid);
         }
-        if self.seq.is_some() {
-            self.seq.map(|s| s + 1);
+        if let Some(seq) = self.seq.as_mut() {
+            *seq += 1;
         }
         if let Some(true) = self.buffer.as_ref().map(|b| b.at_end()) {
             self.buffer = None;
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -354,16 +354,6 @@ impl NlSocket {
     pub fn recv_ack(&mut self) -> Result<(), NlError> {
         if let Ok(ack) = self.recv_nl::<consts::Nlmsg, Nlmsgerr<consts::Nlmsg>>(None) {
             if ack.nl_type == consts::Nlmsg::Error && ack.nl_payload.error == 0 {
-                if self.pid.is_none() {
-                    self.pid = Some(ack.nl_pid);
-                } else if self.pid != Some(ack.nl_pid) {
-                    return Err(NlError::BadPid);
-                }
-                if let Some(seq) = self.seq {
-                    if seq != ack.nl_seq {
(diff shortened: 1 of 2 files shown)


jbaublitz:
Yes, please explain. I typically won't merge code without some sort of justification provided.

vorner:
OK, I'll try to write more than the commit messages, then.

* The first commit removes the seq-check and PID check inside the `recv_ack` message. The very first thing in `recv_ack` is calling `recv_nl`, which already contains the very same check. This is harmless, but also useless and makes the code longer.
* The second commit, I believe the old code was incorrect. It does `self.seq.map(|s| s+1)`. This computes a value one larger, but just returns it ‒ and then throws it away, not stores. Effectively no operation. I guess the original intention was to increment the stored value, not compute and throw away.
* Third, when the kernel sends an error that is not 0 (eg. suggess), and I do `?` in main or anywhere, it previous said „Not ACK“. That is not very useful. This takes the error and looks up the message for it. I guess this is not so great and you mentioned you have plans to do something better eventually, but in the meantime I think this is an improvement.
* The fourth, added later, should do no functional changes, but the old code felt weird to me. So that's just suggestion to make it hopefully more readable, though that is subjective.

If you don't like any of these, I can prepare a branch with just some of the commits.

src/socket.rs
```diff
@@ -332,19 +332,16 @@ impl NlSocket {
...
         if self.pid.is_none() {
             self.pid = Some(msg.nl_pid);
         } else if self.pid != Some(msg.nl_pid) {
             return Err(NlError::BadPid);
         }
-        if self.seq.is_some() {
-            self.seq.map(|s| s + 1);
+        if let Some(seq) = self.seq.as_mut() {
+            *seq += 1;
         }
```
jbaublitz: Yikes. Good catch here.

src/socket.rs
```diff
@@ -354,22 +351,17 @@ impl NlSocket {
...
-                if self.pid.is_none() {
-                    self.pid = Some(ack.nl_pid);
-                } else if self.pid != Some(ack.nl_pid) {
-                    return Err(NlError::BadPid);
-                }
-                if let Some(seq) = self.seq {
-                    if seq != ack.nl_seq {
-                        return Err(NlError::BadSeq);
-                    }
-                }
```
jbaublitz: A lot of the code for PID and ACK handling is very broken and I think it's probably best to remove it for now until I get 0.5.0 released. I'm focused pretty heavily on doing it right there.

src/socket.rs
```diff
@@ -354,22 +351,17 @@ impl NlSocket {
...
                 if let Some(b) = self.buffer.as_mut() {
                     b.rewind()
                 }
-                Err(NlError::NoAck)
+                if ack.nl_type == consts::Nlmsg::Error {
+                    let err = std::io::Error::from_raw_os_error(-ack.nl_payload.error as _);
+                    Err(NlError::Msg(err.to_string()))
+                } else {
+                    Err(NlError::NoAck)
+                }
```
jbaublitz: Looks good.

vorner:
OK, so are there some next steps from me here? (I'm just making sure I'm not overlooking something)

src/socket.rs
```diff
@@ -332,19 +332,16 @@ impl NlSocket {
             mem.truncate(mem_read as usize);
             self.buffer = Some(StreamReadBuffer::new(mem));
         }
-        let msg = match self.buffer {
-            Some(ref mut b) => Nlmsghdr::deserialize(b)?,
-            None => unreachable!(),
-        };
+        let msg = Nlmsghdr::deserialize(self.buffer.as_mut().unwrap())?;
```
jbaublitz: I'd prefer to revert this for readability purposes.

vorner: OK, done.

jbaublitz:
Closed via 73380de64688076d463d6c63cbbcbdb3922feead

Issue #37, opened 2019-09-27
Add support for documentation of enum variants in enum generation macros


Eonm:
Hi,

It is possible to document each enum variant created with the macro impl_var_trait! ?

```rust
    impl_var_trait!(
    /// nl80211Commands
    ///
    /// Enumeration from nl80211/nl80211.h:880
    Nl80211Cmd, u8, Cmd,
    CmdUnspec                  => 0,
    /// Request information about a wiphy or dump request to get a list of all present wiphys.
    /// How can I document this enum variant ?
    CmdGetWiphy                => 1,
    CmdSetWiphy                => 2
    );
```
Thanks

jbaublitz:
Hi @Eonm, I'm going to be reworking the macros as part of #19. I can certainly add support for variant documentation. I don't tend to use it because of the number of enums I support out of the box and I'm not always entirely sure what each one represents due to lack of netlink documentation. However because it's exposed as part of the library, this makes sense to support. I'll edit this issue and will add this feature as I do #19.

Eonm:
Indeed, lack of documentation seems to be the rule with netlink. 

I tried to figure out how to implement this feature myself, but i never used rust macros before.

Thank you for your reply.

jbaublitz:
This should be resolved in commit b00b0bbc7a7975a0cd43d831a40541c63e0e52df. Closing.

PR #48, opened 2019-10-12
Nflog support MVP

vorner:
Hello

This adds the support for basic use of the nflog protocol. There are still some parts missing, but I'd prefer to have a feedback about it sooner than later and I tried to make sure all the missing pieces can be added without breaking API.

I've also tried to structure it in a way that makes it possible/natural to add all the other netfilter protocols over time (NFQUEUE, CONNTRACK).

There are, however, few things I'm not entirely happy about or don't know how to handle properly. I'll probably open separate issues to discuss them one by one, if there's some nice solution for them. But I'll still list my worries here first.

* When receiving messages over the socket, I have to specify into what type I want them to deserialize. But this might depend on the type. This can be seen on the added example, sometimes I get a message with type 3 ‒ which I think corresponds to NLMSG_DONE. I'm not sure if I should make it part of the new `NetfilterMsg` enum, but that would still not help much, because it would still try to deserialize the rest as `LogPacket`. This doesn't currently fail, but is kind of wrong. It would be great if I could somehow specify a nl-type => parse type relation. Unlike `NlAttr<_, Vec<u8>>`, the Nlmsghdr<_, Vec<u8>>` doesn't have methods to parse as one or another type.
* Some of the structs/messages are ever sent in only one direction. So it makes no sense to provide eg. `serialize` for the `LogPacket`. For now, I've put `unimplemented` in them, but it kind of doesn't feel right. Would it make sense to split the Nl trait into reading and writing halves?
* The crate uses the `Vec` type quite generously. For example, when parsing a message, it allocates a Vec as the receive buffer, then it allocates a Vec for all the attributes and, as they have diffferent types, each one is a smaller Vec... As the Vec contains a heap allocation, it might turn somewhat expensive. This probably doesn't matter for things like manipulating the routing table, but with the NFLOG, the number of packets received per second can be quite huge. It would be great if there was a way to use the library with smaller number of allocations (some of my ideas could get down to 0 per packet).
* The `impl_*` macros don't want to let me put documentation onto the individual variants. I've tried to poke at them a little bit, but I've managed to only break existing code, so I've given up for now.

Anyway, if I'm missing anything important here, or if there's any way to improve the code or interface, I'm happy to hear it. I'd then add some fixup commits and rebuild the branch before merging.

### DIFF
--- a/src/consts/netfilter.rs
+++ b/src/consts/netfilter.rs
@@ -1,0 +1,78 @@
+//! Constants for netfilter related protocols
+//!
+//! Note that this doesn't cover everything yet, both the list of types and variants in enums will
+//! be added over time.
+
+use super::{NlAttrType, NlType};
+
+impl_var_trait! {
+    /// Attributes inside a netfilter log packet message.
+    ///
+    /// These are send by the kernel and describe a logged packet.
+    NfLogAttr, u16, NlAttrType,
+    PacketHdr => 1,
+    Mark => 2,
+    Timestamp => 3,
+    IfindexIndev => 4,
+    IfindexOutdev => 5,
+    IfindexPhyindev => 6,
+    IfindexPhyoutdev => 7,
+    Hwaddr => 8,
+    Payload => 9,
+    Prefix => 10,
+    Uid => 11,
+    Seq => 12,
+    SeqGlobal => 13,
+    Gid => 14,
+    Hwtype => 15,
+    Hwheader => 16,
+    Hwlen => 17,
+    Ct => 18,
+    CtInfo => 19
+}
+
+impl_var_trait! {
+    /// Configuration attributes for netfilter logging.
+    ///
+    /// See [LogConfigReq][crate::netfilter::LogConfigReq]
+    NfLogCfg, u16, NlAttrType,
+    Cmd => 1,
+    Mode => 2,
+    NlBufSize => 3,
+    Timeout => 4,
+    QThresh => 5,
+    Flags => 6
+}
+
+impl_var_trait! {
+    /// Messages related to the netfilter netlink protocols.
+    ///
+    /// These appear on the [NlFamily::Netfilter][super::NlFamily::Netfilter] sockets.
+    NetfilterMsg, u16, NlType,
+    // TODO: Docs here /// A logged packet, going from kernel to userspace.
+    LogPacket => 0x0400,
+    // TODO: Docs here /// A logging configuration request, going from userspace to kernel.
+    LogConfig => 0x0401
+}
+
+impl_trait! {
+    /// Parameters for the [NfLogCfg::Cmd].
+    LogCfgCmd, u8
+}
+
+impl_var_trait! {
+    /// Command value for the [NfLogCfg::Cmd].
+    LogCmd, u8, LogCfgCmd,
+    Bind => 1,
+    Unbind => 2,
+    PfBind => 3,
+    PfUnbind => 4
+}
+
+impl_var! {
+    /// Copy mode of the logged packets.
+    LogCopyMode, u8,
+    None => 0,
+    Meta => 1,
+    Packet => 2
+}
--- a/src/netfilter.rs
+++ b/src/netfilter.rs
@@ -1,0 +1,285 @@
+//! Netfilter protocols
+//!
+//! Protocols used for communicating with netfilter. Currently, this contains (partial) support for
+//! NFLOG, NFQUEUE and CONNTRACK will be added later.
+//!
+//! See the examples in the git repository for actual, working code.
+
+use std::ffi::CString;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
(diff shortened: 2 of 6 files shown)


vorner:
Linking here #37, the macros...

vorner:
CI can't compile libc :-(.

vorner:
Note to self. I've just discovered methods like `u32::to_be`. This means all the `htons` are not necessary and I'll get rid of them in a fixup some time soonish.

jbaublitz:
Restart CI here too.

jbaublitz:
CI is fixed.

src/consts/netfilter.rs
```diff
@@ -0,0 +1,78 @@
...
+//! be added over time.
+
+use super::{NlAttrType, NlType};
+
+impl_var_trait! {
+    /// Attributes inside a netfilter log packet message.
+    ///
+    /// These are send by the kernel and describe a logged packet.
+    NfLogAttr, u16, NlAttrType,
+    PacketHdr => 1,
```
jbaublitz: My guess is that these constants don't show up in libc. If we can get them into libc, I'd much prefer that over hard coding constants. We do this elsewhere, but I'd have to see if the libc crate would be open to accepting a PR for that.

src/netfilter.rs
```diff
@@ -0,0 +1,285 @@
...
+
+#[derive(Copy, Clone, Debug)]
+struct Timestamp {
+    secs: u64,
+    usecs: u64,
+}
+
+impl Nl for Timestamp {
+    fn serialize(&self, m: &mut StreamWriteBuffer) -> Result<(), SerError> {
+        u64::to_be(self.secs).serialize(m)?;
```
jbaublitz: Are you positive these are all in big endian? If you are assuming that because netlink is a network protocol, netlink always uses the endianness of the host system, whether big or little, as it is exclusively a local protocol. If you have an example of this protocol specifically being big endian values, can you provide that evidence? [This comment](https://chromium.googlesource.com/native_client/linux-headers-for-nacl/+/2dc04f8190a54defc0d59e693fa6cff3e8a916a9/include/linux/netlink.h#140) was how I originally discovered this as netlink will explicitly let you know when it's network byte order as opposed to host byte order.

src/socket.rs
```diff
@@ -336,9 +336,9 @@ impl NlSocket {
             Some(ref mut b) => Nlmsghdr::deserialize(b)?,
             None => unreachable!(),
         };
-        if self.pid.is_none() {
+        if self.pid.is_none() && msg.nl_pid != 0 {
```
jbaublitz: Please move the socket changes into a separate PR. I'm happy to merge those before this one. I agree that PID tracking is very broken and I'm looking to fix that in v0.5.0 but am happy to turn this off for now in another PR.

vorner: OK, I've submitted the commit as #57. It's the same one, so it should disappear from the diff here once that one is merged into master.

src/netfilter.rs
vorner: I'm positive. First, the type in the headers are marked as `__be32`.

Second, I did it without these conversions first and it didn't work ‒ the values for timestamp, for example, usually produced overflows when converting to SystemTime. When I do the conversion, it produces sane values.

src/consts/netfilter.rs
vorner: I'll open an issue there. Should we wait with this pull request until they decide, or go ahead and update it once a new release is made on their side?

src/netfilter.rs
jbaublitz: Excellent. This looks good then. I know it can get a little tedious providing evidence so thank you for bearing with me.

src/consts/netfilter.rs
jbaublitz: I'll merge this instead of waiting on that because that may take a bit to figure out and they don't always release all that quickly. Once I do merge this, I'm going to open an issue that, if you're willing, I'll assign to you to convert those over to libc constants if that gets merged eventually. Sound good? Also, can you link to the issue you opened here so I can follow up on it? Thanks!

src/socket.rs
jbaublitz: Excellent thank you! I want to think through that one a little more before merging, hence the separate PR. Once we figure that out, I'll ask you to rebase onto master so that you pull in whatever we decide on and then I'll merge this one.

src/consts/netfilter.rs
vorner: Sounds good.

https://github.com/rust-lang/libc/issues/1562

src/netfilter.rs
vorner: That's fine. I take a thorough review as a seal of quality of the library :-).

jbaublitz:
Hi @vorner, thanks for your other contribution. I've merged that PR so can you rebase onto master and force push here? I'll do one final review before merging.

PR #49, opened 2019-10-12
Return unsigned types in NlSocket::{send, recv}

### LINKED ISSUE #47: Should NlSocket::send/recv return unsigned lengths?
Hello

I have noticed that the `NlSocket::recv` and `::send` (the raw/low-level ones) return the amount of data read or written as `ssize_t` (which is usually equivalent to `isize`). This mimics the low-level C API. However, negative values indicate errors and these are indicated as `Result` on the Rust side.

Would it be OK if I send a PR changing them to usize for a more idiomatic API? Should I aim at master or the 0.5-dev branch?

Thank you

vorner:
Return size_t instead of ssize_t. All the negative values of the latter
are considered errors, so never leave the function.

Closes #47.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -192,7 +192,7 @@ impl NlSocket {
 
     /// Send message encoded as byte slice to the netlink ID specified in the netlink header
     /// (`neli::nl::Nlmsghdr`)
-    pub fn send<B>(&mut self, buf: B, flags: i32) -> Result<libc::ssize_t, io::Error>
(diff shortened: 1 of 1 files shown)


vorner:
It seems the CI is ill, but I think it's not anything I've done. It fails to compile libc :-(.

jbaublitz:
I'm going to try fixing CI by rebuilding the container. I'll see if that helps.

jbaublitz:
CI is fixed. Merging.

PR #53, opened 2019-10-15
Document which Rtattr.rta_type goes with what rtnetlink struct.

oherrala:
This information can be found from `rtnetlink(7)` manpage, but small hint in the docs doesn't hurt?


jbaublitz:
Excellent, thanks for the improvements! Merging.

Issue #54, opened 2019-10-15
Is there a reason to keep Rtattr as type variable in structs?


oherrala:
For example struct `Rtmsg` is defined to take type variable `T`. It's now required for the user of the crate to match correct enum with correct struct. And it's easy to make mistakes. But is there really a need for this type variable? I quickly tried the following patch to hard code the `consts::Rta` into `Rtmsg` and it just works. After fixes to `examples/route-list.rs`, the example compiles and works.

```diff
diff --git a/src/rtnl.rs b/src/rtnl.rs
index 7b84dc7..d6d12ac 100644
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -345,7 +345,7 @@ impl Nl for Rtgenmsg {
 
 /// Route message
 #[derive(Debug)]
-pub struct Rtmsg<T> {
+pub struct Rtmsg {
     /// Address family of route
     pub rtm_family: RtAddrFamily,
     /// Length of destination
@@ -365,13 +365,10 @@ pub struct Rtmsg<T> {
     /// Routing flags
     pub rtm_flags: Vec<RtmF>,
     /// Payload of `Rtattr`s
-    pub rtattrs: Rtattrs<T, Vec<u8>>,
+    pub rtattrs: Rtattrs<crate::consts::Rta, Vec<u8>>,
 }
 
-impl<T> Nl for Rtmsg<T>
-where
-    T: RtaType,
-{
+impl Nl for Rtmsg {
     fn serialize(&self, buf: &mut StreamWriteBuffer) -> Result<(), SerError> {
         self.rtm_family.serialize(buf)?;
         self.rtm_dst_len.serialize(buf)?;
@@ -432,7 +429,7 @@ where
                 - rtm_type.size()
                 - mem::size_of::<libc::c_int>(),
         );
-        let rtattrs = Rtattrs::<T, Vec<u8>>::deserialize(buf)?;
+        let rtattrs = Rtattrs::<crate::consts::Rta, Vec<u8>>::deserialize(buf)?;
 
         Ok(Rtmsg {
             rtm_family,
```

oherrala:
I think this has been fixed in #55. Closing.

PR #55, opened 2019-10-15
Simplify structs by removing type variables in rtnl

oherrala:
Rtattrs have specific type depending on the struct. Remove type variable to simplify using these structs and remove possibility to make mistakes.

I made the mistake of using wrong type and spend embarrassingly amount of time figuring out what happens. I opened this PR which implements simpler structs for `rtnl` so that I can give it a more test.

Related to #54.

### DIFF
--- a/examples/route-list.rs
+++ b/examples/route-list.rs
@@ -9,7 +9,7 @@ use neli::nl::Nlmsghdr;
 use neli::rtnl::*;
 use neli::socket::*;
 
-fn parse_route_table(rtm: Nlmsghdr<Rtm, Rtmsg<Rta>>) {
+fn parse_route_table(rtm: Nlmsghdr<Rtm, Rtmsg>) {
     // This sample is only interested in the main table.
     if rtm.nl_payload.rtm_table == RtTable::Main {
         let mut src = None;
@@ -63,7 +63,7 @@ fn parse_route_table(rtm: Nlmsghdr<Rtm, Rtmsg<Rta>>) {
 fn main() -> Result<(), Box<dyn Error>> {
     let mut socket = NlSocket::connect(NlFamily::Route, None, None, true).unwrap();
 
-    let rtmsg: Rtmsg<Rta> = Rtmsg {
+    let rtmsg = Rtmsg {
         rtm_family: RtAddrFamily::Inet,
         rtm_dst_len: 0,
         rtm_src_len: 0,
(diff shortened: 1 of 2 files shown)


jbaublitz:
I need to review this as generic netlink operates very differently. In generic netlink, the type field can be a number of different enums which is why this was added to rtnetlink. I was basing this off of that. It may indeed be the case that this is not necessary for rtnetlink, but I'd like to do some more digging to ensure that this doesn't limit use with other enums if there's ever a case where multiple enums could be used as the type for that field. My initial guess is that this is fine given what I've seen of rtnetlink so far compared to generic netlink, and if you can provide me documentation or some other kind of evidence that these conditions are always met for each of the changed structs, I'll look that over and merge.

oherrala:
I based my PR on the `rtnetlink(7)` manpage (http://man7.org/linux/man-pages/man7/rtnetlink.7.html):

* RTM_NEWLINK, RTM_DELLINK, RTM_GETLINK
  * Attributes listed are prefixed with `IFLA_`
* RTM_NEWADDR, RTM_DELADDR, RTM_GETADDR
  * Attributes listed are prefixed with `IFA_`
* RTM_NEWROUTE, RTM_DELROUTE, RTM_GETROUTE
  * Attributes listed are prefixed with `RTA_`
* RTM_NEWNEIGH, RTM_DELNEIGH, RTM_GETNEIGH
  * Attributes listed are prefixed with `NDA_`
  * This section of manpage is formatted differently than the rest and is missing many attributes which can be found from C header file. See #56 for WIP fix for this message.
* RTM_NEWRULE, RTM_DELRULE, RTM_GETRULE
  * "Carries a struct rtmsg" => attributes same as for `rtmsg`?
* RTM_NEWQDISC, RTM_DELQDISC, RTM_GETQDISC
  * Attributes listed are prefixed with `TCA_`
* RTM_NEWTCLASS, RTM_DELTCLASS, RTM_GETTCLASS
  * "These messages contain a struct `tcmsg` as described above." => attributes same as for `tcmsg`?
* RTM_NEWTFILTER, RTM_DELTFILTER, RTM_GETTFILTER
  * "These messages contain a struct `tcmsg` as described above." => attributes same as for `tcmsg`?


And this stackoverflow answer https://stackoverflow.com/a/25535748/361823 states:

> The "attributes" are protocol dependent, and relate to a specific message being sent using said protocol.

jbaublitz:
I looked at that documentation when writing this and it must not have registered because I started with generic netlink. This is very helpful! Thanks for your contribution. I'm going to merge it.

oherrala:
Thanks @jbaublitz!

PR #57, opened 2019-10-18
Handle unset PID

vorner:


jbaublitz:
I'm looking at this now, but I'm actually considering making `None` indicate that PID checking is disabled.

From what I understand a PID of 0 indicates that the message is being sent to the kernel. Therefore, if it's `None`, ignore the PID checking. If it's `Some(0)`, overwrite it with the PID assigned to the socket when the user binds the socket. This seems like a friendlier way of navigating this problem as `None` really shouldn't be used both as an initialization value and also an indication that PID checking should be used. That's overloading the semantic meaning of the variant.

What do you think?

vorner:
Well, the problems I had were with messages coming *from* the kernel, which had PID 0. But otherwise I have no idea what the purpose of the PID field or the actual check of it is.

I mean, in which case would a wrong message get delivered and the library would be supposed to fail on it?

Anyway, I'm OK with changing it that way. I'll send an updated code some time soon.

vorner:
Added as a fixup. I can either rewrite the branch or you probably can squash the whole branch to a single commit at merge (as it is effectively only one commit).

jbaublitz:
This is great, thanks!

PR #58, opened 2019-10-24
Fix `Ndmsg` struct

### LINKED ISSUE #56: Neighbor table struct (Ndmsg) is broken
The neighbor table struct `Ndmsg` in rtnl.rs is broken:

 * Rtattrs and rta_type for this is missing
 * Padding is missing from (de)serializing
 * Size calculation is wrong due to above

I have been working to fix it (see https://github.com/oherrala/neli/compare/fix-54...oherrala:add-neigh-rtattrs) and it can currently at least deserialize the response from netlink socket, but it's currently on top of #55 and needs to be fixed to be compatible with current master.

oherrala:
As documented in #56, `Ndmsg` struct is broken. This PR fixes it by adding the missing `rtattr` field for the struct, add the missing `rta_type` for `Rtattr`s as well as fixing the alignment of the struct with correct padding. Finally of course fixing the length calculations.

I couldn't find any docs (or use case) for `Ndmsg.ndm_type`, so I changed it to `u8`. It's there, but no idea what it should be used for.

Fixes #56

### DIFF
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -462,7 +462,7 @@ pub struct Ndmsg {
     /// Flags for entry
     pub ndm_flags: Vec<Ntf>,
     /// Type of entry
-    pub ndm_type: u8,
+    pub ndm_type: Rtn,
     /// Payload of `Rtattr`s
     pub rtattrs: Rtattrs<Nda, Vec<u8>>,
 }
@@ -526,7 +526,7 @@ impl Nl for Ndmsg {
             }
             ndm_flags
         };
-        let ndm_type = u8::deserialize(buf)?;
+        let ndm_type = Rtn::deserialize(buf)?;
 
         buf.set_size_hint(
             size_hint
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -535,7 +535,7 @@ impl Nl for Ndmsg {
                 - ndm_index.size()
                 - mem::size_of::<u16>() // ndm_state
                 - mem::size_of::<u8>() // ndm_flags
-                - ndm_type.size()
+                - ndm_type.size(),
         );
 
         let rtattrs = Rtattrs::<Nda, Vec<u8>>::deserialize(buf)?;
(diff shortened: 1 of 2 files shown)


jbaublitz:
@oherrala I'm going to request leaving `.ndm_type` as is based on some source code that uses the type I specified. I didn't see any documentation but apparently `RTN_*` constants are valid per [this code](https://github.com/FRRouting/frr/blob/master/nhrpd/netlink_arp.c#L47).

src/rtnl.rs
```diff
@@ -462,12 +462,16 @@ pub struct Ndmsg {
     /// Flags for entry
     pub ndm_flags: Vec<Ntf>,
     /// Type of entry
-    pub ndm_type: Rtn,
+    pub ndm_type: u8,
```
jbaublitz: Please change this back to `Rtn`.

oherrala:
@jbaublitz I think you are right. I did more research and found for example this Linux patch:

  https://lore.kernel.org/patchwork/patch/487044/

which was merged and it confirms the usage of `RTN_` in `ndm_type`. I'll fix my PR.

The funny thing is that the latest `rtnetlink.7` manpage doesn't document `ndm_type` (as seen here http://man7.org/linux/man-pages/man7/rtnetlink.7.html), but there still exist some old version of man page (for example http://www.skrenta.com/rt/man/rtnetlink.4.html) which document it:

> Valid ndm_type values are: NDA_UNSPEC  for  unknown  type, NDA_DST  for  a  neighbour cache network layer destination address,  ND_LLADDR  for  a  neighbour  cache  link  layer address,  ND_CACHEINFO  for  cache  statistics  (a  struct nda_cacheinfo  header   follows).

However, the current Git project of man-pages project doesn't contain the change to remove this part of documentation so it has happened before 1.70 man-pages release (that's where the current Git repo starts: https://github.com/mkerrisk/man-pages/commit/fea681dafb1363a154b7fc6d59baa83d2a9ebc5c).

I filed a bug into Linux's bugzilla: https://bugzilla.kernel.org/show_bug.cgi?id=205345

Maybe the documentation can be improved.

PR #61, opened 2019-11-21
Extend RtAddrFamily and allow changing change mask

marcelbuesing:
I'm working on [exchanging](https://github.com/mbr/socketcan-rs/pull/13) the netlink-rs dependency in the socketcan-rs crate for neli but need this two changes to make it work. 
Adding `set_ifi_change` instead of specifying it in the new fn should make this a nonbreaking change.

### DIFF
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -167,7 +167,7 @@ impl Ifinfomsg {
             ifi_type,
             ifi_index,
             ifi_flags: vec![Iff::Up],
-            ifi_change: Iff::Up.into(),
+            ifi_change: Iff::Up,
             rtattrs,
         }
     }
@@ -184,7 +184,7 @@ impl Ifinfomsg {
             ifi_type,
             ifi_index,
             ifi_flags: Vec::new(),
-            ifi_change: Iff::Up.into(),
+            ifi_change: Iff::Up,
             rtattrs,
         }
     }
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -130,7 +130,8 @@ pub struct Ifinfomsg {
     pub ifi_index: libc::c_int,
     /// Interface flags
     pub ifi_flags: Vec<Iff>,
-    ifi_change: libc::c_uint,
+    /// Change mask
+    pub ifi_change: libc::c_uint,
     /// Payload of `Rtattr`s
     pub rtattrs: Rtattrs<Ifla, Vec<u8>>,
 }
@@ -152,6 +153,11 @@ impl Ifinfomsg {
             ifi_change: 0xffff_ffff,
             rtattrs,
         }
+    }
+
+    /// Set change mask
+    pub fn set_ifi_change(&mut self, ifi_change: libc::c_uint) {
+        self.ifi_change = ifi_change;
     }
 }
 
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -155,9 +155,38 @@ impl Ifinfomsg {
         }
     }
 
-    /// Set change mask
-    pub fn set_ifi_change(&mut self, ifi_change: libc::c_uint) {
-        self.ifi_change = ifi_change;
+    /// Set the link with the given index up (equivalent to `ip link set dev DEV up`)
+    pub fn up(
+        ifi_family: RtAddrFamily,
+        ifi_type: Arphrd,
+        ifi_index: libc::c_int,
+        rtattrs: Rtattrs<Ifla, Vec<u8>>,
+    ) -> Self {
+        Ifinfomsg {
+            ifi_family,
+            ifi_type,
+            ifi_index,
+            ifi_flags: vec![Iff::Up],
+            ifi_change: Iff::Up.into(),
+            rtattrs,
+        }
+    }
+
+    /// Set the link with the given index down (equivalent to `ip link set dev DEV down`)
+    pub fn down(
+        ifi_family: RtAddrFamily,
+        ifi_type: Arphrd,
+        ifi_index: libc::c_int,
+        rtattrs: Rtattrs<Ifla, Vec<u8>>,
+    ) -> Self {
+        Ifinfomsg {
+            ifi_family,
+            ifi_type,
+            ifi_index,
+            ifi_flags: Vec::new(),
+            ifi_change: Iff::Up.into(),
+            rtattrs,
+        }
     }
 }
 
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -131,7 +131,7 @@ pub struct Ifinfomsg {
     /// Interface flags
     pub ifi_flags: Vec<Iff>,
     /// Change mask
-    pub ifi_change: libc::c_uint,
+    pub ifi_change: Iff,
     /// Payload of `Rtattr`s
     pub rtattrs: Rtattrs<Ifla, Vec<u8>>,
 }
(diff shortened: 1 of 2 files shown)


jbaublitz:
Hi @marcelbuesing! Thanks for letting me know about this. The documentation states that that field is for future use but it is entirely possible that this assertion has gone stale. Can you explain to me a little bit about what this field does? I'd be interested in making it more intuitive to new users particularly in the case that it is a bitmask (there's already some work on the `v0.5.0` branch to make that simpler so I'd love to include this in those changes for v0.5). I also would recommend the following to avoid a breaking change. I'm definitely okay with making the field public but I want new to remain as is for now (as you have done). What I'd prefer to a `set` method is a second `new` method with a different name that includes the new parameter. I will consolidate the two in v0.5 as breaking changes are acceptable there. As this release is taking a lot longer than I'd like, I'll also release a new version in the v0.4 range so that you can use this update in the meantime. Thanks!

little-dude:
The `ifi_change` flag is used when toggling a property off I think. For instance, in my netlink crate, to set a link UP and DOWN [I have the following](https://github.com/little-dude/netlink/blob/252822f8c72e497bfcbe7794eac963b089d9b7ad/rtnetlink/src/link/set.rs#L53):

```rust
    /// Set the link with the given index up (equivalent to `ip link set dev DEV up`)
    pub fn up(mut self) -> Self {
        self.message.header.flags = IFF_UP;
        self.message.header.change_mask = IFF_UP;
        self
    }

    /// Set the link with the given index down (equivalent to `ip link set dev DEV down`)
    pub fn down(mut self) -> Self {
        self.message.header.change_mask = IFF_UP;
        self
    }
```

`pyroute2` does the same, but to give another example, here is how they toggle [arp on and off](https://github.com/svinota/pyroute2/blob/d119a60bd36ad1af30d7512d1af2cbd1e24226ab/pyroute2/iproute/linux.py#L1312-L1319):

```python
        # arp on/off shortcut
        if 'arp' in kwarg:
            mask |= IFF_NOARP
            if not kwarg.pop('arp'):
                flags |= IFF_NOARP

        msg['flags'] = flags
        msg['change'] = mask
```

FInally [in the `netlink` go package](https://github.com/vishvananda/netlink/blob/d71301a47b607450337d920f260f3dc76481298e/link_linux.go#L301-L315) they have:

```go
// LinkSetDown disables link device.
// Equivalent to: `ip link set $link down`
func (h *Handle) LinkSetDown(link Link) error {
	base := link.Attrs()
	h.ensureIndex(base)
	req := h.newNetlinkRequest(unix.RTM_NEWLINK, unix.NLM_F_ACK)

	msg := nl.NewIfInfomsg(unix.AF_UNSPEC)
	msg.Change = unix.IFF_UP
	msg.Index = int32(base.Index)
	req.AddData(msg)

	_, err := req.Execute(unix.NETLINK_ROUTE, 0)
	return err
}
```

Apparently, it can also be used when listening for netlink event, to detect when a flag changed.
This is a little less clear to me because I haven't used it but here are a couple links:

- https://amnonpaz.com/2016/09/14/using-rtnetlink-for-detection-of-nics-status-changes-in-c/
- https://stackoverflow.com/a/28405729/1836144

Digging a little in the linux source code it seems that whenever the flags are changed with (`dev_change_flags`), and update is multicasted with `dev_notify_flags`.

```c
int dev_change_flags(struct net_device *dev, unsigned int flags,
		     struct netlink_ext_ack *extack)
{
	int ret;
	unsigned int changes, old_flags = dev->flags, old_gflags = dev->gflags;

	ret = __dev_change_flags(dev, flags, extack);
	if (ret < 0)
		return ret;

	changes = (old_flags ^ dev->flags) | (old_gflags ^ dev->gflags);
	__dev_notify_flags(dev, old_flags, changes);
	return ret;
}
```
However the code in `__dev_notify_flags` is a little complex so I stopped digging.

marcelbuesing:
@little-dude thanks for all the research, I also found this:

> Flags Change Mask (32bit) The primary use of this field is to specify a mask of flags that should be changed based on the value of the Flags field. A special meaning is given to this field when present in link notifications, see TODO.
https://www.infradead.org/~tgr/libnl/doc/route.html

> Finally after a lot of research I was able to get this done. There is a flag called "ifi_change" which gives out the change in the Netdevice's state change. I wasn't able to think of it because the rtnetlink man page did not mention anything about it. It says it is reserved for future use and should always be set to 0xFFFFFFFF. However, its value changes upon change in the current state of the interface. If there is a state change it gives a finite value else it is zero. Using this check I was able to suppress multiple netlink messages.
https://stackoverflow.com/a/28405729/1045920

ip link using the flag for bringing an interface up or down:
https://github.com/shemminger/iproute2/blob/afa588490b7e87c5adfb05d5163074e20b6ff14a/ip/iplink.c#L599

Although making this more typesafe using bitflags would be a great idea it seems difficult to determine the possible values given the state of documentation ...

little-dude:
> it seems difficult to determine the possible values given the state of documentation

the possible values are basically the `IFF_*` constants or `0xffff_ffff`

marcelbuesing:
> > it seems difficult to determine the possible values given the state of documentation
> 
> the possible values are basically the `IFF_*` constants or `0xffff_ffff`

Sounds like a good idea to me.

marcelbuesing:
Anything I can still do here ?

src/rtnl.rs
```diff
@@ -153,6 +154,40 @@ impl Ifinfomsg {
...
+        ifi_type: Arphrd,
+        ifi_index: libc::c_int,
+        rtattrs: Rtattrs<Ifla, Vec<u8>>,
+    ) -> Self {
+        Ifinfomsg {
+            ifi_family,
+            ifi_type,
+            ifi_index,
+            ifi_flags: Vec::new(),
+            ifi_change: Iff::Up.into(),
```
jbaublitz: This looks like a copy paste error.

src/rtnl.rs
```diff
@@ -130,7 +130,8 @@ pub struct Ifinfomsg {
     pub ifi_index: libc::c_int,
     /// Interface flags
     pub ifi_flags: Vec<Iff>,
-    ifi_change: libc::c_uint,
+    /// Change mask
+    pub ifi_change: libc::c_uint,
```
jbaublitz: I think it's best to convert this to the type `Iff` for now. I have a way to mark multiple types as usable in a field in the `consts` infrastructure if someone comes along and requests that other values should be supported. Until then, I think it's safe to assume that `Iff` is a reasonable accepted type.

jbaublitz:
@marcelbuesing Are you okay with me making the requested edits on top of your commits and merging? I'd like to get this in.

marcelbuesing:
Sure please feel free to change everything!

src/rtnl.rs
marcelbuesing: Actually let me briefly check this one it might not be a copy and paste error. I remember being surprised here too.

marcelbuesing: Yes this is meant to be like this you can also see this in @little-dudes comment show e.g. the golang version https://github.com/jbaublitz/neli/pull/61#issuecomment-557836682

marcelbuesing:
Rebased and changed the type as requested to Iff.

src/rtnl.rs
marcelbuesing: Done

jbaublitz:
See #77

PR #65, opened 2020-01-01
Ignore Cargo.lock

gluxon:
Hey John, I just learned today that `Cargo.lock` shouldn't be committed and published with library crates.

See https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries


jbaublitz:
Good catch! I learned something. Merging

PR #66, opened 2020-01-13
NlSocket::from_raw_fd

vorner:
Implement the FromRawFd trait on the socket, to complement the IntoRawFd
and AsRawFd.

My motivation is that I want to create the socket in one process and send it over the unix domain socket to another process.

I have a question about what to do with the seq field. The current code kind of ignores it, so it disables the tracking. What is a possible way forward? Add a `set_seq_track` onto the socket? Possibly with setting the socket number too?


jbaublitz:
See #76

PR #67, opened 2020-01-15
Handle missing padding after last attribute

vorner:
The requirement is for each attribute to *start* on word boundary, but
not necessarily end, it seems to be legal for the last attribute in a
message (and buffer) not to be followed by a padding. Apparently at
least the NFQUEUE takes advantage and sends such messages.

This relaxes the strictness about that and silently just assumes the end
of buffer was reached.

I'm not 100% sure all three changes are needed, I was discovering them one by one by the means of hunting errors and panics.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -128,7 +128,9 @@ pub trait Nl: Sized {
     {
         let padding_len = self.asize() - self.size();
         if padding_len > 0 {
-            m.read_exact(&mut [0; libc::NLA_ALIGNTO as usize][..padding_len])?;
+            // Ignore result - the padding is mandatory between attributes, but not after the last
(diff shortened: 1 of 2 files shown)


jbaublitz:
This is how I've set up genetlink. Are you simply making it consistent with what I've done in genetlink?

vorner:
I don't know what you did with genetlink, I did this because I'm experimenting with creating support for the NFQUEUE protocol. I have it in a branch, but I need to clean it up before submitting.

Without these changes, it failed to parse many messages, because the last attribute is usually the packet payload ‒ it has variable length. And someone of the kernel devs wanted to save the additional bytes in the padding and I was getting parse errors.

PR #70, opened 2020-01-29
Fix the nflog mark

vorner:
* All marks in iptables/netfilter are 32 bit everywhere.
* According to experiments, it seems to be big endian.

(Sorry for all these little fixes, I didn't get it right the first time because the netlink/netfilter documentation is somewhat lacking)

### DIFF
--- a/src/netfilter.rs
+++ b/src/netfilter.rs
@@ -59,7 +59,7 @@ pub struct LogPacket {
     ///
     /// A mark used through the netfilter, working as kind of scratch memory. 0 and no mark set are
     /// considered equivalent.
-    pub mark: u16,
+    pub mark: u32,
     /// A timestamp when the packet has been captured.
     pub timestamp: SystemTime,
     /// Source hardware address (eg. MAC).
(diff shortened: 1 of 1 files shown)


jbaublitz:
I'm going to review all of your PRs now that I'm back from vacation. Could you please provide me with a few detailed sets of commands that I can use to test this manually?

vorner:
```
$ iptables -I INPUT -j NFLOG --nflog-group 10
$ iptables -I INPUT -j MARK --set-mark 0x10203040
```

Then run the nflog example from the repo. It'd print all the packets, including the mark (decadic, so you need to convert them to verify).

It should start printing messages like:

```
LogPacket { hw_protocol: 512, hook: 0, mark: 270544960, timestamp: SystemTime { tv_sec: 0, tv_nsec: 0 }, hwaddr: [144, 92, 68, 250, 33, 35], payload: [69, 0, 0, 52, 3, 116, 64, 0, 55, 6, 118, 254, 151, 101, 112, 133, 192, 168, 0, 191, 1, 187, 211, 90, 156, 99, 108, 163, 85, 233, 230, 193, 128, 16, 0, 60, 213, 48, 0, 0, 1, 1, 8, 10, 58, 232, 158, 89, 76, 175], prefix: "", ifindex_in: Some(2), ifindex_out: None, ifindex_physin: None, ifindex_physout: None, uid: Some(1000), gid: Some(1000), attr_len: 156 }
```

(See the mark field)

PR #75, opened 2020-03-15
Update tokio to 0.2

jbaublitz:
@another-s347 I've made changes to your PR but this should update everything to tokio 0.2. I'm in the process of reimplementing sink, but let me know if you'd like to see anything else before I merge.


another-s347:
Everything looks good.

PR #76, opened 2020-03-15
Remove IntoRawFd

jbaublitz:
PR revisions for #66 

@vorner I really don't think `IntoRawFd` should have been implemented in the first place. I really don't feel good about leaking IO resources so I'm going to remove it altogether. `AsRawFd` should do almost exactly the same thing without breaking resource cleanup. Is there a reason you were looking for this functionality? If so could you provide me with an example?

Either way, thank you for bringing this to my attention!

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -26,7 +26,7 @@ use std::{
     io,
     marker::PhantomData,
     mem::{size_of, zeroed},
-    os::unix::io::{AsRawFd, IntoRawFd, RawFd},
+    os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd},
 };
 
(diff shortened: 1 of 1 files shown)


vorner:
Yes, there was a specific reason. Due to some tests I needed to create the socket in one process and pass it to another process through a unix domain socket (it's possible to send sockets from one process to another). So I'd do IntoRawFd in one process, send it and do FromRawSocket in the other.

I've also tried to use the netlink protocol over something that wasn't a netlink socket and the library helped me with that, by injecting some other socket.

I know this is kind of stuff is out of ordinary, but there are such needs from time to time. These three functions allow many crazy things and interoperability with other languages.

jbaublitz:
Okay, I'm going to experiment with using `std::mem::forget` here instead. I'll put something up for you to review.

PR #77, opened 2020-03-15
Fix for Ifinfomsg

jbaublitz:
@marcelbuesing I made one addition to your PR and rebased it. Take a look and let me know if you want to discuss anything else before I merge. Upon further reflection, convenience methods for commonly generated packets may make this library a whole lot more usable. Thank you for your contribution!


marcelbuesing:
Looks good to me !

jbaublitz:
I've changed the target to 0.5.0 so I'm removing the `new2` method in favor of changing the new method. Merging.

PR #92, opened 2020-09-05
rtnl: Add example to get all routable IPs

joshtriplett:
This example also demonstrates how to read a netlink message
generically, and interpret differently depending on its type, using
`StreamReadBuffer` and `Nl::deserialize`.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -21,6 +21,10 @@ name = "genl_stream"
 path = "genl_stream.rs"
 
 [[bin]]
+name = "getips"
+path = "getips.rs"
+
+[[bin]]
 name = "route-list"
 path = "route-list.rs"
 
--- a/examples/getips.rs
+++ b/examples/getips.rs
@@ -1,0 +1,62 @@
+extern crate neli;
+
+use std::net::Ipv4Addr;
+
+use neli::attr::Attribute;
+use neli::consts::nl::{NlmF, NlmFFlags};
+use neli::consts::rtnl::{Ifa, IfaFFlags, RtAddrFamily, RtScope, Rtm};
+use neli::consts::socket::NlFamily;
+use neli::err::NlError;
+use neli::nl::{NlPayload, Nlmsghdr};
+use neli::rtnl::Ifaddrmsg;
+use neli::socket::NlSocketHandle;
+use neli::types::RtBuffer;
+use neli::utils::U32Bitmask;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut rtnl = NlSocketHandle::connect(NlFamily::Route, None, U32Bitmask::empty())?;
+    let ifaddrmsg = Ifaddrmsg {
+        ifa_family: RtAddrFamily::Inet,
+        ifa_prefixlen: 0,
+        ifa_flags: IfaFFlags::empty(),
+        ifa_scope: 0,
+        ifa_index: 0,
+        rtattrs: RtBuffer::new(),
+    };
+    let nl_header = Nlmsghdr::new(
+        None,
+        Rtm::Getaddr,
+        NlmFFlags::new(&[NlmF::Request, NlmF::Root]),
+        None,
+        None,
+        NlPayload::Payload(ifaddrmsg),
+    );
+    rtnl.send(nl_header)?;
+    let mut addrs = Vec::<Ipv4Addr>::with_capacity(1);
+    for response in rtnl.iter(false) {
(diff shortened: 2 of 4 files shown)


jbaublitz:
Thanks for the additional example! I'll try to take a look soon.

jbaublitz:
@joshtriplett I've been really trying to get this next release out. I haven't forgotten about this! I'm releasing in a few minutes and then I can plan to get around to this again.

jbaublitz:
@joshtriplett Would you be willing to rework this for 0.5.0 now that it's out?

joshtriplett:
@jbaublitz In the course of forward-porting this, I ran into a bug in the definition of `Ifaddrmsg`: the serialization and deserialization tries to use `IfaFFlags`, which is a 4-byte value, but the flags field in the message is a 1-byte field. (Messages that use the larger `IFA_F_*` values provide an attribute containing 4-byte flags that supersedes the 1-byte flags field in the initial part of the message.) The serialization and deserialization of Ifaddrmsg need to treat that as a 1-byte field. (Ideally, serialization should produce an error for excessively large values rather than silently truncating.)

joshtriplett:
I've confirmed that changing the type of `ifa_flags` to `libc::c_uchar` allows parsing an `Ifaddrmsg`.

It's *possible* to fix this without a major version bump, by serializing and deserializing `Ifaddrmsg` field by field. I can provide a patch for that. But I think it might make sense to use two different types for `IfaF`: one for the short 1-byte version used in the initial part of `Ifaddrmsg`, and another for the extended 4-byte version that can appear as an attribute. (neli could provide a `From` impl to convert from the 1-byte version to the 4-byte version, and a `TryFrom` impl to convert from the 4-byte version to the 1-byte version.)

joshtriplett:
Also, this issue aside, I love the new API. Being able to just iterate, and have neli handle Multi/Done message series for me, works really well. The code got substantially shorter and clearer.

joshtriplett:
I've rewritten the code to compile with 0.5. This should not be merged until #109 is fixed, though.

jbaublitz:
> It's possible to fix this without a major version bump, by serializing and deserializing Ifaddrmsg field by field. I can provide a patch for that. But I think it might make sense to use two different types for IfaF: one for the short 1-byte version used in the initial part of Ifaddrmsg, and another for the extended 4-byte version that can appear as an attribute. (neli could provide a From impl to convert from the 1-byte version to the 4-byte version, and a TryFrom impl to convert from the 4-byte version to the 1-byte version.)

Can you file a separate issue for breaking these out into separate types? I'll target this in 0.6.0 and add that milestone so I don't lose track of it. For now, I think the best thing I can do here is your suggestion that doesn't break backwards compatibility. There is actually a way to accomplish this with minimal changes that still does not require serializing field by field. I can convert IfaFFlags to an integer, do a pre-check for serialization, returns an error if the provided flag value will be too large, and then pass the correctly-sized integer as the serialization field in the macro. For deserialization, I need to look into this a little bit more carefully. This may have to be done field by field. Does this sound reasonable to you?

> Also, this issue aside, I love the new API. Being able to just iterate, and have neli handle Multi/Done message series for me, works really well. The code got substantially shorter and clearer.

I'm so glad this is helpful! I saw that as a major usability gap.

One additional note is that based on #102, I'm trying to add more infrastructure to test actually using neli to interact with the kernel. Can you tell me a little bit more about how you found this bug and what the error was in issue #102? I'd like to add a test so that I can make sure this doesn't happen again.

joshtriplett:
> Can you file a separate issue for breaking these out into separate types? I'll target this in 0.6.0 and add that milestone so I don't lose track of it.

Done in #111.

> For now, I think the best thing I can do here is your suggestion that doesn't break backwards compatibility. There is actually a way to accomplish this with minimal changes that still does not require serializing field by field. I can convert IfaFFlags to an integer, do a pre-check for serialization, returns an error if the provided flag value will be too large, and then pass the correctly-sized integer as the serialization field in the macro. For deserialization, I need to look into this a little bit more carefully. This may have to be done field by field. Does this sound reasonable to you?

That sounds reasonable for 0.5, to avoid breaking compatibility.

> Can you tell me a little bit more about how you found this bug and what the error was in issue #102? I'd like to add a test so that I can make sure this doesn't happen again.

For this particular case, this example serves as the test case: if run without fixing the handling of `ifa_flags` in `Ifaddrmsg`, it produces an error: `Error: Msg("The buffer was not large enough to complete the deserialize operation")`

I'll post some more ideas in #102.

Issue #86, opened 2020-05-03
Make all neli struct fields private and generate/implement getters for fields


jbaublitz:
I'm toying with the idea of exposing all packet creation through convenience methods like `new()` and others as mentioned in #83 and autogenerating getters for all of the fields in the packets. This would provide a few things:
* Less confusion on how to construct packets. All of the methods will be documented.
* Getters will be available to read the fields for use with responses that have been deserialized so that fields are no longer public.

jbaublitz:
@ollipa Given that this is somewhat unrelated to the builder pattern and more related to the visibility of struct fields after the builder has been used to build those structs, I'm going to count this as separate and leave this open as orthogonal work.

jbaublitz:
Closed by #182

Issue #83, opened 2020-04-12
Add convience methods and builder pattern for neli struct construction


jbaublitz:
Currently, constructing larger and more complicated packets to be sent over netlink is not necessarily intuitive. Add some convenience functions for common use cases to avoid having to specify every field each time a header is created.

jbaublitz:
@ollipa Do you mind if I add you to assignees here just to keep track of the fact that we're working on this together?

ollipa:
Sure add me to the assignees :ok_hand:

jbaublitz:
See PR #93 for a more in depth discussion.

PR #93, opened 2020-11-22
Implement Nlmsghdr builder

ollipa:
This PR adds builder for `Nlmsghdr` struct so that it would be a bit more ergonomic to create netlink packets. If you like the idea I'd be happy to implement something similar for other types as well.

This PR doesn't contain any breaking changes but I made some changes to `Nlmsghdr` based on your your comments in issues #83 and #86. You can find those changes in this commit https://github.com/ollipa/neli/commit/91843b3860a5f40ff899a2767bd0f2d0568b4adb (note that the code doesn't compile). I wonder if this is the way you were planning to change the API.

### DIFF
--- a/src/nl.rs
+++ b/src/nl.rs
@@ -18,6 +18,87 @@ use crate::{
     Nl,
 };
 
+/// Builder for [Nlmsghdr]
+/// # Examples
+///
+/// ```
+/// # use neli::consts::{NlmF, Nlmsg};
+/// # use neli::nl::NlEmpty;
+/// # use neli::nl::NlmsghdrBuilder;
+/// let nl = NlmsghdrBuilder::new(Nlmsg::Noop, NlEmpty)
+///     .nl_flag(NlmF::Dump)
+///     .build();
+/// ```
+#[derive(Debug, PartialEq)]
+pub struct NlmsghdrBuilder<T, P> {
+    nl_len: Option<u32>,
+    nl_type: T,
+    nl_flags: Vec<NlmF>,
+    nl_seq: Option<u32>,
+    nl_pid: Option<u32>,
+    nl_payload: P,
+}
+
+impl<T, P> NlmsghdrBuilder<T, P>
+where
+    T: NlType,
+    P: Nl,
+{
+    /// Create a new top level netlink packet builder
+    pub fn new(nl_type: T, nl_payload: P) -> Self {
+        Self {
+            nl_len: None,
+            nl_type,
+            nl_flags: Vec::new(),
+            nl_seq: None,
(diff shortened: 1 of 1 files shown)


jbaublitz:
While I'm not opposed to the builder pattern, I'm not sure it's the best pattern for neli. I'm still considering it, but given that so many of the arguments are required, I don't see a huge benefit in the builder pattern over a `new` method for all of the structs that need them. While it is idiomatic in Rust, I see the main advantage being for structs that have a lot of optional arguments. If I eventually move in the direction of more optional arguments, I would be far more open to the builder pattern. What are your thoughts?

ollipa:
The way I see it, builder would make the API nicer to use and the code more readable especially when there are multiple `Option`s passed to the struct like in `Nlmsghdr`. Maybe it would make sense to remove `nl_len`, `nl_seq`, and `nl_pid` from the constructor and provide setter methods for them instead?

Anyway, feel free to close this PR if you are not planning to add builders yet. Also, thank you for your work with the crate! I'm going to test out v0.5 soon and looking forward for more improvements :slightly_smiling_face:

jbaublitz:
I'm not opposed to discussing this, but before I'd be willing to merge this, I'd have to see a case for this being applicable for the other structs in neli too. I can imagine a few different patterns that could be useful in this context and when I release v0.6.0, I just want to make sure that the pattern is uniform. For example, I would need to look at the `rtnl.rs` module in particular to see how many of those structs have reasonable defaults for fields. If you can make the case for most of the structs in neli having reasonable defaults for at least some of the fields, this could absolutely be an improvement, but if not, I'd prefer to take the approach of covering the majority of common defaults each with an appropriately named method such as [this method](https://www.infradead.org/~tgr/libnl/doc/api/group__send__recv.html#gafb5bcaba9d349cf699e4c8ce2c72aab8) in `libnl` but for message construction as opposed to sending. Providing a set of methods that fill in defaults for common use cases like that method may also be an option and I want to make sure that we discuss it in more detail before deciding. Given that this is targeting v0.6.0, it may be a while until I make a decision, but I'd love to keep chatting about it until then!

ollipa:
Okay, got it! It is definitely a good idea to keep the API consistent. I'll update my project to neli 0.5 and then try to think how builder pattern (or some alternative approach) could work with other structs.

jbaublitz:
Perfect! Please do keep me posted because if we can implement a consistent builder pattern, that could absolutely be an improvement. I'd love to work with you on this so feel free to update me with any future ideas or proposals in issue #83. I think that's the appropriate place to follow up on any planning.

PR #103, opened 2020-12-05
Example for custom generic netlink family with custom attributes

phip1611:
I'd like to add an example on **how to use neli with custom properties for custom generic netlink families**. I use this at an uni project where I develop a kernel module (which registers a custom generic netlink family), a userland component in Rust (using neli), and a other userland component using libnl. Therefore I can ensure that my approach works with the kernel / other netlink implementations.

Let's assume we have a custom kernel driver that registers a `"your_family_name_here"` family. The subject of our problem knows the operations/commands `echo`, `create device`, and `delete device`. It also knows the attributes `msg (string)` and `device name (string)`. 

In this example (see code) an `echo` message is send to the kernel and the reply received:

_PS: Example kernel module that fulfills this contract: https://github.com/phip1611/generic-netlink-user-kernel-rust_ 

PPS: When I started doing this project (with neli 0.4.3) I spend so much time on this. Therefore I'd love to have an easy example in this great lib so that others can profit.

Let me know what you think :)

### DIFF
--- a/examples/custom_generic_nl_family_custom_types.rs
+++ b/examples/custom_generic_nl_family_custom_types.rs
@@ -23,7 +23,7 @@ use neli::{
     types::{Buffer, GenlBuffer},
     utils::U32Bitmask,
 };
-use std::{process, process::exit};
+use std::process;
 
 /// Name of the Netlink family registered via Generic Netlink
 const FAMILY_NAME: &str = "gnl_foobar_xmpl";
@@ -73,9 +73,9 @@ fn main() {
                 "The Netlink family '{}' can't be found. Is the kernel module loaded yet? neli-error='{}'",
                 FAMILY_NAME, e
             );
-            // exit without error in order for Continuous Integration and automatic testing not to fail
-            // when the kernel module is not loaded
-            exit(0);
+            // Exit without error in order for Continuous Integration and automatic testing not to fail.
+            // This is because in testing/build scenarios we do not have a Kernel module which we can load.
+            return;
         }
     }
 
--- a/examples/custom_generic_nl_family_custom_types.rs
+++ b/examples/custom_generic_nl_family_custom_types.rs
@@ -2,7 +2,7 @@
 //! family via Generic Netlink. The family is called "gnl_foobar_xmpl" and the
 //! kernel module must be loaded first. Otherwise the family doesn't exist.
 //!
-//! A working kernel module implementation, with that you can use this binary,
+//! A working kernel module implementation with which you can use this binary
 //! can be found here: https://github.com/phip1611/generic-netlink-user-kernel-rust
 //!
 //! Output might look like this (if the kernel module is loaded)
@@ -100,8 +100,8 @@ fn main() {
     //    attributes (actual data) as payload.
     let gnmsghdr = Genlmsghdr::new(
         NlFoobarXmplOperation::Echo,
-        // You can evolve your application over time using different versions or ignore it
-        // Application specific; receiver must take care of it
+        // You can evolve your application over time using different versions or ignore it.
+        // Application specific; receiver can check this value and to specific logic
         1,
         // actual payload
         attrs,
@@ -111,10 +111,14 @@ fn main() {
     let nlmsghdr = Nlmsghdr::new(
         None,
         family_id,
-        // This depends on the receiving side. Do you check there if any flags are present?
-        // Request-flag is required (TODO by neli or by netlink?)
-        // others are up to you
+        // You can use flags in an application specific way (e.g. ACK flag). It is up to you
+        // if you check against flags in your Kernel module. It is required to add NLM_F_REQUEST,
+        // otherwise the Kernel doesn't route the packet to the right Netlink callback handler
+        // in your Kernel module. This might result in a deadlock on the socket if an expected
+        // reply you are waiting for in a blocking way is never received.
+        // Kernel reference: https://elixir.bootlin.com/linux/v5.10.16/source/net/netlink/af_netlink.c#L2487
         NlmFFlags::new(&[NlmF::Request]),
+        // It is up to you if you want to split a data transfer into multiple sequences. (application specific)
         None,
         // Port ID. Not necessarily the process id of the current process. This field
         // could be used to identify different points or threads inside your application
--- a/examples/custom_generic_nl_family_custom_types.rs
+++ b/examples/custom_generic_nl_family_custom_types.rs
@@ -2,7 +2,7 @@
 //! family via Generic Netlink. The family is called "gnl_foobar_xmpl" and the
 //! kernel module must be loaded first. Otherwise the family doesn't exist.
 //!
-//! A working kernel module implementation, where you can use this binary,
+//! A working kernel module implementation, with that you can use this binary,
 //! can be found here: https://github.com/phip1611/generic-netlink-user-kernel-rust
 //!
 //! Output might look like this (if the kernel module is loaded)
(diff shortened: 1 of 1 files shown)


phip1611:
Of course this will not build/terminate in any CI system.

examples/custom_generic_nl_family_custom_types.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    // Send data
+    sock.send(nlmsghdr).expect("Send must work");
+
+    // receive echo'ed message
+    let res: Nlmsghdr<u16, Genlmsghdr<FoobarOperation, FoobarAttribute>> =
+        sock.recv().expect("Should receive a message").unwrap();
+
+    // According to Netlink Linux kernel code nl_type is either 0x2 (error) or our family id (ok)
+    if res.nl_type == 0x2 {
```
phip1611, line 127: Ah I have to refactor this. Looks like this check is already covered by sock.recv(). It returns Err if nl_type is 0x2

phip1611: No, neli tells `The buffer was not large enough to complete the deserialize operation`. Doesn't look what I would neli expect to do. Technically an "well formed error message" is okay but not error

phip1611:
Probably found a general neli problem that is not related to this specific PR. I just want to note it:
If I reply from kernel with a netlink error message (`nl_type` in netlink-header is `NLMSG_ERR (0x2)`) then ´socket.recv()´ returns `ERR` because of `The buffer was not large enough to complete the deserialize operation`. I don't know if this is right. A well formed error message from the kernel should be okay. Netlinks man page also encourages developers to use NLMSG_ERR to signalize error. `.recv()` should only return Err if it couldn't parse the buffer.

jbaublitz:
@phip1611 Thanks for the report! So just to confirm, you returned an error packet with no `nlmsgerr` payload? According to the netlink manpage, this does not appear to be valid behavior at this point in time.

```NLMSG_ERROR message signals an error and the payload contains an nlmsgerr structure```

from [here](https://www.man7.org/linux/man-pages/man7/netlink.7.html).

Unless you can give me a specific link to documentation that specifies that this payload is optional, I think this is a problem in your implementation, not in mine.

**Comment by jbaublitz** [comment]
I am happy to create an issue to improve error messaging if that would be helpful for you. I've also bumped into some situations where the error messaging is not adequate and I'd like to improve the error provided to the user in certain cases.

**Comment by phip1611** [comment]
@jbaublitz You are right. I'm not a netlink expert and I just tried to use it in a way that it works for me. My Kernel module didn't include `struct nlmsgerr` in the payload. But can you imagine a more convenient way to recognize the absence of this struct, when `nl_type` is `NLMSG_ERR`? Because the current message _"The buffer was not large enough to complete the deserialize operation"_ is not effective.

PS: I update my kernel module to include the struct in the payload in error case. I edit this comment when I have an update on this.

**Comment by jbaublitz** [comment]
Perfect! I'll open an issue and tag you in it so we can discuss details.

**Comment by jbaublitz** [comment]
I have a few comments on the code so far. I'll review this soon and request some changes just so that people looking at the code are guided towards neli best practices when they look at the example.

File: examples/custom_generic_nl_family_custom_types.rs [hunk 2]
```diff
@@ -0,0 +1,138 @@
...
+use neli::consts::nl::Nlmsg;
+use neli::consts::nl::{NlmF, NlmFFlags};
+use neli::consts::socket::NlFamily;
+use neli::genl::{Genlmsghdr, Nlattr};
+use neli::nl::{NlPayload, Nlmsghdr};
+use neli::socket::NlSocketHandle;
+use neli::types::{Buffer, GenlBuffer};
+use neli::utils::U32Bitmask;
+use neli::Nl;
+use std::process;
```
jbaublitz, line 21: Could you switch this to nested imports? I've been meaning to put out some formal coding style guidelines but haven't gotten around to that quite yet.

examples/custom_generic_nl_family_custom_types.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    // Send data
+    sock.send(nlmsghdr).expect("Send must work");
+
+    // receive echo'ed message
+    let res: Nlmsghdr<u16, Genlmsghdr<FoobarOperation, FoobarAttribute>> =
+        sock.recv().expect("Should receive a message").unwrap();
+
+    // According to Netlink Linux kernel code nl_type is either 0x2 (error) or our family id (ok)
+    if res.nl_type == u16::from(Nlmsg::Error)
```
jbaublitz, line 123: neli has special handling for the `Nlmsg::Error` value as of 0.5.0. This value indicates either an actual error if the error code is non-zero or an ACK if the error code is 0. neli will return these as values of either `NlPayload::Ack` or `NlPayload::Err` in the resulting `Nlmsghdr`. In the error case, `recv()` will detect this as an error and return it as an `Err(Nlmsgerr)` value. Can you explain a little bit more why you chose to do the checking manually here? Does this use case not work for you?

examples/custom_generic_nl_family_custom_types.rs
```diff
@@ -0,0 +1,138 @@
...
+        // depends on your specific use case
+        panic!("Kernel replied with error");
+    }
+    if res.nl_type == family_id {
+        println!("Received successful reply!");
+    }
+
+    let attr_handle = res.get_payload().unwrap().get_attr_handle();
+    let received = attr_handle.get_attribute(FoobarAttribute::Msg).unwrap();
+    let received = String::deserialize(received.nla_payload.as_ref()).unwrap();
```
jbaublitz, line 136: This can be more concisely represented by `attr_handle.get_attr_payload_as::<String>(FoobarAttribute::Msg)`.

jbaublitz (changes requested):
Here are some initial notes.

examples/custom_generic_nl_family_custom_types.rs
phip1611: Oh, apparently I was using the neli lib wrong in my uni project and therefore also in this demo. 
I didn't had a proper example at the first place that fit my use case, therefore I just tried to figure everything out by myself. I'll improve the code and let you know if it works better. @jbaublitz

phip1611:
@jbaublitz FYI I'll fix this in the next weeks, I didn't forget it. I'm having a deadline for an uni project this sunday (where I use neli :))

phip1611:
@jbaublitz Sorry it took a while. Any more comments on the code? Of course I tested it and it's working.

jbaublitz:
I just reran the failed test and realized that this is a new clippy error in my code. I'll fix that.

jbaublitz:
@phip1611 Clippy checks are now optional in CI. If you rebase onto master, you should be good to go.

phip1611:
@jbaublitz  done; pushed the rebased branch

examples/custom_generic_nl_family_custom_types.rs
```diff
@@ -0,0 +1,146 @@
...
+    match res {
+        Ok(id) => family_id = id,
+        Err(e) => {
+            eprintln!(
+                "The Netlink family '{}' can't be found. Is the kernel module loaded yet? neli-error='{}'",
+                FAMILY_NAME, e
+            );
+            // exit without error in order for Continuous Integration and automatic testing not to fail
+            // when the kernel module is not loaded
+            exit(0);
```
jbaublitz, line 78: This can just be `return`

jbaublitz (changes requested):
Once this last change is made, I'll merge this. Thanks for putting so much effort into the documentation!

phip1611:
@jbaublitz I fixed it.

jbaublitz (approved):

PR #110, opened 2020-12-22
WIP: Improve error messaging

### LINKED ISSUE #105: Improve error messaging
Created due to #103 

@phip1611 I'm envisioning the following for the `SerError` and `DeError` structs which generated the error that was confusing for you. Because a lot of these errors are autogenerated in the `serialize` and `deserialize` macros, how does this sound? `SerError` and `DeError` will include the type of the object being serialized or deserialized, an optional struct field if it's a struct being serialized or deserialized, and what the cause of the error was. So an error in the case of your example would look something like this:

```Deserializing field error of struct Nlmsgerr could not be completed: the end of the deserialization buffer was reached before the deserialization operation could be completed; the packet received may be incomplete```

Here I'm assuming that you included no payload whatsoever after the netlink header so deserialization would have failed on the deserialization of `Nlmsgerr` [here on this line](https://github.com/jbaublitz/neli/blob/master/src/nl.rs#L214).

Would that error message be clearer?

jbaublitz:
Closes #105

### DIFF
--- a/src/err.rs
+++ b/src/err.rs
@@ -263,7 +263,9 @@ impl Display for SerError {
 
 impl Error for SerError {}
 
-/// Deserialization error
+/// Deserialization error. Most variants contain fields for the
+/// type and an optional struct field where deserialization failed.
+// TODO: Remove deprecated variants in 0.6.0.
 #[derive(Debug)]
 pub enum DeError {
     /// Abitrary error message.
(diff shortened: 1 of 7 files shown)


phip1611:
@jbaublitz I've tested your branch in my uni project. It shows a few more information now which is great! It helps me to further investigate an issue. I still have to clarify whether the error is in the kernel or in neli - but I have a few more information now. Thanks!

jbaublitz:
@phip1611 Hello! I've been on vacation for the past two weeks so I'm picking back up on this. Did you figure out which it is?

@joshtriplett This is still a WIP so that is not surprising. :) I'm updating the title to reflect that and will make sure to resolve that before it's merged.

phip1611:
The error is/was in my kernel module. So neli works fine. @jbaublitz

PR #112, opened 2021-01-03
Fix size of ifa_flags field in Ifaddrmsg

### LINKED ISSUE #109: Ifaddrmsg serialization/deserialization uses the wrong size for ifa_flags
In the course of forward-porting https://github.com/jbaublitz/neli/pull/92 , I ran into a bug in the definition of `Ifaddrmsg`: the serialization and deserialization tries to use `IfaFFlags`, which is a 4-byte value, but the flags field in the message is a 1-byte field. (Messages that use the larger `IFA_F_*` values provide an attribute containing 4-byte flags that supersedes the 1-byte flags field in the initial part of the message.) The serialization and deserialization of `Ifaddrmsg` need to treat that as a 1-byte field. (Ideally, serialization should produce an error for excessively large values rather than silently truncating.)

Looks like this was introduced in a53abe0 as part of the switch from `Vec<IfaF>` to `IfaFFlags`.

I've confirmed that changing the type of `ifa_flags` to `libc::c_uchar` allows parsing an `Ifaddrmsg`. (That's not the right type, but it's the right size, so I used it for testing purposes.)

It's possible to fix this without a major version bump, by serializing and deserializing `Ifaddrmsg` field by field. I can provide a patch for that. But I think it might make sense to use two different types for `IfaF`: one for the short 1-byte version used in the initial part of `Ifaddrmsg`, and another for the extended 4-byte version that can appear as an attribute. (neli could provide a `From` impl to convert from the 1-byte version to the 4-byte version, and a `TryFrom` impl to convert from the 4-byte version to the 1-byte version.)

joshtriplett:
Ifaddrmsg currently uses IfaFFlags for its ifa_flags field, but the
structure on the wire only has one-byte flags, not four-byte flags.
(Four-byte flags get sent as an attr.)

Fix the serialization and deserialization of Ifaddrmsg, Add From and
TryFrom impls to convert IfaFFlags to and from libc::c_uchar.

Closes #109

### DIFF
--- a/src/consts/rtnl.rs
+++ b/src/consts/rtnl.rs
@@ -372,10 +372,37 @@ impl_flags!(
     #[allow(missing_docs)]
     pub IffFlags, Iff, libc::c_uint
 );
+
 impl_flags!(
     #[allow(missing_docs)]
     pub IfaFFlags, IfaF, libc::c_uint
 );
+
+impl std::convert::TryFrom<&IfaFFlags> for libc::c_uchar {
+    type Error = std::num::TryFromIntError;
+    fn try_from(flags: &IfaFFlags) -> Result<libc::c_uchar, Self::Error> {
+        let mut n: libc::c_uint = 0;
+        for bit in 0..std::mem::size_of::<libc::c_uint>() * 8 {
+            if flags.contains(&IfaF::from(1 << bit)) {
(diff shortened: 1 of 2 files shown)


jbaublitz (approved):
Okay this looks great! Merging.

PR #118, opened 2021-01-05
Ifinfomsg: Remove duplicate serialization of ifi_change field

joshtriplett:
Ifinfomsg was serializing two copies of ifi_change; drop the incorrect
one.


joshtriplett:
Ran into this when testing some code using `Rtm::Getlink`. This affects sending any message containing an `Ifinfomsg`.

jbaublitz (approved):
Good catch, thank you!

Issue #101, opened 2020-12-05
Re-evaluate serde


jbaublitz:
serde might be an option for a layer underneath the `Nl` trait to replace or use beside the macros for serialization and deserialization. This could be beneficial for users that want to serialize a JSON representation of netlink packets for logging, etc.

jbaublitz:
@joshtriplett I've been a bit absent on GIthub mainly because I've been working on this issue for a while. I know you had previously asked why I was storing flags as a vector and I am beginning work on fixing that up in my serde implementation. I wanted to ask what you see as the best design choice here. The options I see are:

1. allocate a slice that has a size large enough to hold all flags
2. keep it as a `Vec`
3. store it as an integer representation

3 is probably the best option for obvious reasons but it will take a little bit of special handling (using `TryFrom` instead of `From` for example to make sure the Flags passed in don't overflow the integer in the case of the `UnrecognizedConst` variant).

jbaublitz:
I'm beginning to think that serde won't be flexible enough to handle this. The ultimate problem here is that serde, to some extent, doesn't seem to be targeted towards something like network packet parsing. netlink has quite a bit of conditional parsing that varies widely from struct to struct (for example, the payload type differing depending on whether the packet type is an error or not, the payload of the packet that caused the error only being returned only if the error code is non-zero and not for ACKs). serde seems to be targeted at creating a unified serializer and deserializer for a standard data-type based format and I don't think netlink works very well with this paradigm. For example, there is no way, without context, in netlink to differentiate between a `u32` that is a packet length and a `u32` that is simply a value. With serde, it is very hard to communicate this additional context to the deserializer because all it is seeing is a `u32` with no additional information.

I've posted a longer issue [here](https://github.com/serde-rs/serde/issues/2030) and for now, I'm changing focus to using nom which seems like it is a little bit more targeted towards my use case.

PR #127, opened 2021-02-01
Add Ndmsg::new function to initialize Ndmsg

oherrala:
Commit a53abe0ed416040b5edd9f7c5e0d5f05fbab6eb5 introduced two non-public fields to `Ndmsg` struct (`pad1` and `pad2`). This has made it impossible to construct `Ndmsg` struct outside `neli` from version 0.5 and later. This PR introduces `Ndmsg::new()` method to construct the struct.

### DIFF
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -401,6 +401,29 @@ pub struct Ndmsg {
     pub rtattrs: RtBuffer<Nda, Buffer>,
 }
 
+impl Ndmsg {
+    /// Create a fully initialized neighbor table struct
+    pub fn new(
+        ndm_family: RtAddrFamily,
+        ndm_index: libc::c_int,
+        ndm_state: NudFlags,
+        ndm_flags: NtfFlags,
+        ndm_type: Rtn,
+        rtattrs: RtBuffer<Nda, Buffer>,
+    ) -> Self {
+        Ndmsg {
+            ndm_family,
+            pad1: 0,
+            pad2: 0,
+            ndm_index,
+            ndm_state,
+            ndm_flags,
+            ndm_type,
+            rtattrs,
+        }
+    }
+}
+
 impl Nl for Ndmsg {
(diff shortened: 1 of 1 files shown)


jbaublitz:
I'll merge this and include this in the next release. Before I do, I'm wondering if it's worth investigating in #101 whether I can find a way to serialize appropriate alignment for netlink. My understanding based on what I've seen is that netlink needs every four byte increment to be aligned so if you have two `u16`s, that aligns to the four byte boundary. If you have a `u8` and then a `u32`, this would require padding the `u8` with three bytes to preserve 4-byte alignment for both fields. What I'm curious about is whether a `u8`, `u16`, and then `u8` would require padding. There are two different alignment issues at play here: the compiler's x86_64 alignment and netlink's alignment. The x86_64 alignment, as I understand it, would align the `u8` with one byte of padding, then the `u16` would have no padding, and then following `u8` would also have one byte of padding. Because this is not 4-byte aligned, my guess is that this struct:

```rust
struct PaddingExample {
    field1: u8,
    field2: u16,
    field3: u8,
}
```

would:
1. not exist in netlink code. I think this would instead be written in the order `field1`, `field3`, and `field2`.
2. require two bytes of padding added at then end (a field like `padding1: u16`) if it were written this way.

@joshtriplett Based on your experience with kernel alignment, does this sound right to you? @oherrala I'm also happy to hear any feedback you have on this as well.

jbaublitz:
For now I'm going to merge this so that you can get moving @oherrala.

jbaublitz (approved):

joshtriplett:
@jbaublitz That structure wouldn't typically exist in the kernel either, without rearranging its fields for more natural alignment.

Are you saying that netlink just needs the end of the struct padded out to a multiple of 4 bytes, or that each field needs to be?

If the former, that's something you can get the Rust compiler to do, by using `repr(C, align(4))`. The size of a `repr(C, align(N))` struct is always extended to a multiple of `N`.

Issue #134, opened 2021-02-15
nl80211 does not always sends data.


Kloenk:
Some nl80211 commands does only return a Ack, and no data messages. This breaks the current assumptions of NlSocketHandle.

PR #135, opened 2021-02-15
NlSocketHandle: do not expect another ack message, if only an ack message is send

Kloenk:
fix for #134

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -565,6 +565,12 @@ impl NlSocketHandle {
             return Err(NlError::Nlmsgerr(e));
         }
 
+        // return ack message if no other messages are present
+        if let NlPayload::Ack(_) = packet.nl_payload {
+            self.needs_ack = false;
+            return Ok(Some(packet));
+        }
+
         if self.needs_ack
             && (!packet.nl_flags.contains(&NlmF::Multi)
                 || packet.nl_type.into() == Nlmsg::Done.into())


Kloenk:
I guess the failing clippy pipeline is not from my changes.

jbaublitz:
@Kloenk Two comments:
1. The clippy failure is indeed not your fault. Don't worry about that one.
2. I don't actually return the ACKs but do validation for them in the `NlSocketHandle`. Your change would cause the `ACK` to be returned which I don't do anywhere else. I would prefer to pass back an `Ok(None)` after the ACK validation is done.

Can you tell me a little bit more about why you want the ACK passed back?

Kloenk:
Mostly for sequence validation. But I could also work with just Ok(None). Only error ist really annoying

jbaublitz:
If I put up a PR, can you test it today?

jbaublitz:
Also just a note that if you ever want to parse the ACK yourself, `recv_all()` does not processing and just passes back all of the parsed packets in a sequence.

Kloenk:
> If I put up a PR, can you test it today? I'm looking to get the next release out today.

Yes, I can do that. Should I help somehow?

jbaublitz:
Just give #138 a try and let me know how that works.

Kloenk:
On it, will probably take around an hour, as I have to compile a kernel to test it (This is only because I have a unsigned regdb, and does not change the nl80211 interface)

jbaublitz:
Not a problem! Let me know once you've verified it. Closing in favor of #138.

PR #138, opened 2021-02-21
Do not assume that the packet will not be an ACK in NlSocketHandle::recv

### LINKED ISSUE #134: nl80211 does not always sends data.

jbaublitz:
Closes #134

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -581,30 +581,39 @@ impl NlSocketHandle {
         log!("Message received: {:#?}", packet);
 
         if let NlPayload::Err(e) = packet.nl_payload {
-            return Err(NlError::Nlmsgerr(e));
-        }
-
-        if self.needs_ack
-            && (!packet.nl_flags.contains(&NlmF::Multi)
-                || packet.nl_type.into() == Nlmsg::Done.into())
-        {
-            let is_blocking = self.is_blocking()?;
-            self.nonblock()?;
-            self.needs_ack = false;
-            let potential_ack = self.recv::<T, P>()?;
-            if let Some(NlPayload::Payload(_))
-            | Some(NlPayload::Empty)
-            | Some(NlPayload::Err(_))
-            | None = potential_ack.as_ref().map(|p| &p.nl_payload)
+            Err(NlError::from(e))
+        } else if let NlPayload::Ack(_) = packet.nl_payload {
+            if self.needs_ack {
+                self.needs_ack = false;
+                Ok(None)
+            } else {
+                Err(NlError::new(
+                    "Socket did not expect an ACK but one was received",
+                ))
+            }
+        } else {
+            if self.needs_ack
+                && (!packet.nl_flags.contains(&NlmF::Multi)
+                    || packet.nl_type.into() == Nlmsg::Done.into())
             {
-                return Err(NlError::NoAck);
+                let is_blocking = self.is_blocking()?;
+                self.nonblock()?;
+                self.needs_ack = false;
+                let potential_ack = self.recv::<T, P>()?;
+                if let Some(NlPayload::Payload(_))
+                | Some(NlPayload::Empty)
+                | Some(NlPayload::Err(_))
+                | None = potential_ack.as_ref().map(|p| &p.nl_payload)
+                {
+                    return Err(NlError::NoAck);
+                }
+                if is_blocking {
+                    self.block()?;
+                }
             }
-            if is_blocking {
-                self.block()?;
-            }
-        }
 
-        Ok(Some(packet))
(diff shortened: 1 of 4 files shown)


src/socket.rs
```diff
@@ -562,30 +562,39 @@ impl NlSocketHandle {
...
-            {
-                return Err(NlError::NoAck);
+            Err(NlError::from(e))
+        } else if let NlPayload::Ack(_) = packet.nl_payload {
+            if self.needs_ack {
+                self.needs_ack = false;
+                Ok(None)
+            } else {
+                Err(NlError::new(
+                    "Socket did not expect an ACK but one was received",
```
Kloenk, line 572: Shouldn't this be something like NlError::NoAckWanted

Kloenk:
This breaks `socket.resolve_genl_family("nl80211")?`. It creates "netlink error: Socket did not expect an ACK but one was received"

Kloenk:
Just setting `self.needs_ack = true;` after [this line](https://github.com/jbaublitz/neli/blob/issue-134/src/socket.rs#L371) should probably fix it

src/socket.rs
jbaublitz: It should be, but that would be a breaking change for anyone matching on the enum. I'll put up an issue for that in 0.6.0.

jbaublitz:
> Just setting `self.needs_ack = true;` after [this line](https://github.com/jbaublitz/neli/blob/issue-134/src/socket.rs#L371) should probably fix it

I'm going to dig into this further because that should already be set in the implementation of `NlSocketHandle::send()` so obviously something is broken.

Kloenk:
I thing it's only in send_nl and not in send. I think send is a more low level api

jbaublitz:
`send_nl` has been renamed to `send` in `NlSocketHandler` so it should be set.

Kloenk:
Maybe this is there, because it does not use recv, but iter? Not sure, not used the iter function yet, just saw that the resolve family function uses it.

jbaublitz:
@Kloenk I'm sorry for the delay. I've primarily been working on #143. I'm shifting focus back to this and I wanted to propose something.

I'm counting the current implementation as a bug and so I'm willing to introduce a breaking change in a patch version if the current solution does not work.

What I'm currently considering is:
* `NlSocketHandle::recv` should only return `None` if the socket is non-blocking
* `NlMessageIter` should handle the calculation of `None` for the user
* `NlSockHandle::recv` should return _all_ packets to the user without filtering out ACKS (I will even eventually need this for #104)

How does that sound to you?

Kloenk:
I never worked with `NlMessageIter`, so don't have an opinion on that. What would be the return type for `NlSockHanle::recv`? As far as the return type is easily handable, I'm completely fine with your proposal.

jbaublitz:
The return type would remain the same.

Kloenk:
What if the kernel sends an Ack and a Payload? Do you have to to call `recv` twice, or how would you design this?

jbaublitz:
You would have to call `.recv()` twice and to simply iterate until the end, you would use `NlMessageIter`.

jbaublitz:
@Kloenk Once you give this the okay, I'll merge! I've tested it extensively and think this should work now.

src/socket.rs
```diff
@@ -394,21 +394,22 @@ impl NlSocketHandle {
     /// Convenience function for resolving a [`str`] containing the
     /// generic netlink family name to a numeric generic netlink ID.
     pub fn resolve_genl_family(&mut self, family_name: &str) -> Result<u16, NlError> {
+        let mut res = Err(NlError::new(format!(
+            "Generic netlink family {} was not found",
```
Kloenk, line 398: I like to do stuff like this in `'{}'`. but I guess that's a personal favourite, I'm fine with either way

Kloenk (approved):
Did not test it, but so far LGTM

jbaublitz:
@Kloenk Okay sounds good! Let me know when you test it and then I'll merge and put out the new release.

jbaublitz:
Or if you want me to test it, just tell me what packet I should send to the `nl80211` family.

Kloenk:
Will test it right now, but have to build a new kernel (old build is not cached anymore and the nl80211 stuff I'm testing uses EXPERT features of the kernel)

Kloenk:
The tests in my project are succeeding. So works for my current needs. 
LGTM

will this mean a release today? (asking if I should commit the version from git or wait for a tag on crates.io)

jbaublitz:
This does mean a release today. I'll do it right when I get home.

PR #144, opened 2021-03-13
Allow room in error structure for payload as a buffer to be passed back

### LINKED ISSUE #116: Socket buffer should clear itself after deserialize failure happened, because afterwards all operations fail
I use neli custom generic netlink families. I'm pretty sure that this time the bug is inside neli and not my kernel module. I'm sending an error message from the kernel. `nl_type` is `NLMSG_ERR (0x2)` and the payload is only `struct nlmsgerr`.

(If I send regular messages everthing works. I can send thousands of messages in a loop without bugs. It's about the struct nlmsgerr in this case.)

I'm pretty sure that I'm doing it correctly, since I used original Linux code as reference.

The first error message is recognized perfectly fine. But after this point all "ok" and error messages are also rejected with the same problem.  It says `Deserializing type Nlmsghdr could not be completed: the deserialization operation completed before all of the data in the buffer could be consumed; the packet being deserialized may be of the wrong type or an incorrectly sized deserialization buffer may have been provided to the deserialization operation`.

**What I did in userland**
I send a generic netlink message to the kernel which forces an error response.

**Two suggestions for improvements:**

- it would be really useful if https://github.com/jbaublitz/neli/issues/105 would also tell at which offset it failed. For example `nlmsg_len does not match the buffer: 4 bytes are missing` / is too long/too small or so. Currentl I can't figure out where neli fails. It tells me it can't deserialize struct Nlmsghdr.. but why? What bytes are missing?
- After a deserialize failure the socket buffer should be resetted / zeroed, so that future operations will not also fail (or some corresponding internal state; depending on neli implementation)

The code I'm using to send nlmsg_error struct:
```C
void ocp_reply_with_error(struct genl_info *info, int err_code) {
    struct sk_buff *reply_skb;
    int rc;
    struct nlmsghdr *nlmsghdr;
    struct nlmsgerr *nlmsgerr;
    size_t payload_size;

    // netlink standard expects a negative error code.
    // at this point it can happen that we may receive
    // the positive value. Therefore we negate it.

    if (err_code > 0) {
        err_code = -err_code;
    }

    // that's all the payload that we send
    payload_size = sizeof(struct nlmsgerr);

    // just allocates memory
    reply_skb = nlmsg_new(NLMSG_GOODSIZE, GFP_KERNEL);;
    if (reply_skb == NULL) {
        opr_err("ocp_nlmsg_new() failed\n");
        return;
    }

    // We don't send a generic netlink message (generic netlink header after netlink header)
    // we only send a netlink message here!

    // Code similar to https://elixir.bootlin.com/linux/v5.4.86/source/net/netlink/af_netlink.c#L2382

    // puts attributes into netlink header and returns the pointer to the
    // beginning of the netlink header (inside the socket buffer)
    nlmsghdr = nlmsg_put(reply_skb, info->snd_portid, 0, NLMSG_ERROR, payload_size, 0);
    if (nlmsghdr == NULL) {
        opr_err("genlmsg_put() failed because of ENOMEM\n");
        return;
    }

    // returns pointer to head of message payload
    nlmsgerr = nlmsg_data(nlmsghdr);
    // negative status code
    nlmsgerr->error = err_code;
    // netlink header that caused the problem
    // memcpy(&nlmsgerr->msg, info->nlhdr, info->nlhdr->nlmsg_len);
    // for now I only append the pure netlink header of the request and no further data
    // (like gen netlink header or attributes)
    memcpy(&nlmsgerr->msg, info->nlhdr, payload_size);

    // make sure that no userland lib wants to read more data than is available as of now
    // so far we only send the header and not the original attributes
    nlmsgerr->msg.nlmsg_len = sizeof(struct nlmsghdr);

    // update size in netlink header
    nlmsg_end(reply_skb, nlmsghdr);

    // Send the message back to the sender
    rc = nlmsg_unicast(genl_info_net(info)->genl_sock, reply_skb, info->snd_portid);

    if (rc != 0) {
        opr_err("genlmsg_reply() failed because of %d\n", rc);
        return;
    }
}
```

jbaublitz:
Closes #116 

@phip1611 Does this resolve #116 for you?

### DIFF
--- a/src/err.rs
+++ b/src/err.rs
@@ -164,7 +164,9 @@ where
             mem;
             Nlmsgerr::<T> {
                 error: libc::c_int,
-                nlmsg: NlmsghdrErr<T>
+                nlmsg: NlmsghdrErr<T> => {
+                    mem.len() - size_of::<libc::c_int>()
+                },
             }
         })
     }
--- a/src/err.rs
+++ b/src/err.rs
@@ -17,12 +17,14 @@
 use std::{
     error::Error,
     fmt::{self, Debug, Display},
-    io, str, string,
+    io,
+    mem::size_of,
+    str, string,
 };
 
 use crate::{
-    consts::nl::{NlType, NlTypeWrapper, NlmFFlags},
-    types::{DeBuffer, SerBuffer},
+    consts::nl::{NlType, NlmFFlags},
+    types::{Buffer, DeBuffer, SerBuffer},
     Nl,
 };
 
@@ -40,6 +42,40 @@ pub struct NlmsghdrErr<T> {
     pub nl_seq: u32,
     /// ID of the netlink destination for requests and source for responses
     pub nl_pid: u32,
+    nl_payload: Buffer,
+}
+
+impl<T> NlmsghdrErr<T>
+where
+    T: NlType,
+{
+    /// Create a new error packet payload.
+    pub fn new(
+        nl_len: u32,
+        nl_type: T,
+        nl_flags: NlmFFlags,
+        nl_seq: u32,
+        nl_pid: u32,
+        nl_payload: Option<Buffer>,
+    ) -> Self {
+        NlmsghdrErr {
+            nl_len,
+            nl_type,
+            nl_flags,
+            nl_seq,
+            nl_pid,
+            nl_payload: nl_payload.unwrap_or_default(),
+        }
+    }
+
+    /// Deserialize the payload received in the error packet as
+    /// type parameter `P`.
+    pub fn get_payload_as<P>(&self) -> Result<P, DeError>
+    where
+        P: Nl,
+    {
+        P::deserialize(self.nl_payload.as_ref())
+    }
 }
 
 impl<T> Nl for NlmsghdrErr<T>
@@ -53,7 +89,8 @@ where
             self.nl_type;
             self.nl_flags;
             self.nl_seq;
-            self.nl_pid
+            self.nl_pid;
+            self.nl_payload
         }
         Ok(())
     }
@@ -66,7 +103,14 @@ where
                 nl_type: T,
                 nl_flags: NlmFFlags,
                 nl_seq: u32,
-                nl_pid: u32
+                nl_pid: u32,
+                nl_payload: Buffer => {
+                    (nl_len as usize).checked_sub(
+                        size_of::<u32>() * 3
+                      + T::type_size().expect("Constant size")
+                      + NlmFFlags::type_size().expect("Constant size")
+                    ).ok_or(DeError::UnexpectedEOB)?
+                },
             }
         })
     }
@@ -77,14 +121,11 @@ where
             + self.nl_flags.size()
             + self.nl_seq.size()
             + self.nl_pid.size()
+            + self.nl_payload.size()
     }
 
     fn type_size() -> Option<usize> {
-        Some(
-            u32::type_size().expect("Must be constant size") * 3
-                + T::type_size().expect("Must be constant size")
-                + NlmFFlags::type_size().expect("Must be constant size"),
-        )
+        None
     }
 }
 
@@ -156,7 +197,7 @@ pub enum NlError {
     /// Variant for [`String`]-based messages.
     Msg(String),
     /// An error packet sent back by netlink.
-    Nlmsgerr(Nlmsgerr<NlTypeWrapper>),
+    Nlmsgerr(Nlmsgerr<u16>),
     /// A serialization error.
     Ser(SerError),
     /// A deserialization error.
@@ -176,7 +217,7 @@ pub enum NlError {
 
 err_from!(
     NlError,
-    Nlmsgerr<NlTypeWrapper> { NlError::Nlmsgerr },
+    Nlmsgerr<u16> { NlError::Nlmsgerr },
     SerError { NlError::Ser },
     DeError { NlError::De },
     WrappedError { NlError::Wrapped },
--- a/src/err.rs
+++ b/src/err.rs
@@ -42,6 +42,11 @@ pub struct NlmsghdrErr<T> {
     pub nl_seq: u32,
     /// ID of the netlink destination for requests and source for responses
     pub nl_pid: u32,
+    /// Optionally the payload of the packet that caused the error
+    ///
+    /// For ACKs, this will be an empty vector. For errors, this
+    /// will be the serialized payload of the packet that caused
+    /// the error.
     nl_payload: Buffer,
 }
 
--- a/src/err.rs
+++ b/src/err.rs
@@ -105,7 +105,7 @@ where
                 nl_seq: u32,
                 nl_pid: u32,
                 nl_payload: Buffer => {
-                    (nl_len as usize).checked_sub(
+                    mem.len().checked_sub(
                         size_of::<u32>() * 3
                       + T::type_size().expect("Constant size")
                       + NlmFFlags::type_size().expect("Constant size")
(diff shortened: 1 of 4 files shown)


phip1611 (commented):
The Doc Comment of `NlmsghdrErr` should no longer say "with no paylaod.

jbaublitz:
Ah! Thanks so much for catching that! I'll change that and then is this good to merge for you?

phip1611:
Unfortunately your commit breaks my code. Now I get the error: `Type Nlmsgerr<u16> has no static size associated with it`


This can be reproduced using this code:

```rust
    let mut sock = NlSocketHandle::connect(
        NlFamily::Generic,
        // 0 is pid of kernel -> socket is connected to kernel
        Some(0),
        U32Bitmask::empty(),
    )
    .unwrap();

    // FAILS HERE at unwrap
    sock.resolve_genl_family(FAMILY_NAME).unwrap();
```

Therefore I can't test it yet. Any idea why this happens?

PS: The next man page release will contain the updated information what the structure of `struct nlmsgerr` actually is/can be.

jbaublitz:
> Unfortunately your commit breaks my code. Now I get the error: `Type Nlmsgerr<u16> has no static size associated with it`
> 

Whoops! That was a bug on my end. This kind of thing should be resolved in the future with both a migration to serde (#101) and the subsequent removal of these macros.

> This can be reproduced using this minimal code:
> 
> ```rust
>     let mut sock = NlSocketHandle::connect(
>         NlFamily::Generic,
>         // 0 is pid of kernel -> socket is connected to kernel
>         Some(0),
>         U32Bitmask::empty(),
>     )
>     .unwrap();
> 
>     // FAILS HERE at unwrap
>     sock.resolve_genl_family(FAMILY_NAME).unwrap();
> ```
> 
> Therefore I can't test it yet. Any idea why this happens? The Netlink family I use definitely exists, i.e. the code snippet works with neli@0.5.2.
> 
> PS: The next man page release will contain the updated information what the structure of `struct nlmsgerr` actually is/can be.

Thanks so much for getting that resolved!

src/err.rs
```diff
@@ -40,6 +42,40 @@ pub struct NlmsghdrErr<T> {
     pub nl_seq: u32,
     /// ID of the netlink destination for requests and source for responses
     pub nl_pid: u32,
+    nl_payload: Buffer,
```
phip1611, line 45: Maybe you can add a convenient doc comment, like: `Optionally the payload of the original message that caused the error.`?

phip1611:
With 0943bb4 the error is now `Type NlmsghdrErr<T> has no static size associated with it`  @jbaublitz

phip1611:
Thanks for working on this! It looks good now. FYI: This is the code I use to extract the error:

https://github.com/phip1611/generic-netlink-user-kernel-rust/commit/6adc6430a60dbb280441566c72178b4b3cf440c2

Or just the relevant snippet:

```rust
    let res: Result<
        Option<Nlmsghdr<u16, Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>>>,
        NlError,
    > = sock.recv();
    let received_err = res.unwrap_err();
    let received_err = match received_err {
        NlError::Nlmsgerr(err) => err,
        _ => panic!("We expected an error here!"),
    };
    println!(
        "Inside the kernel this error code occurred: {}",
        received_err.error
    );
    let nlmsg_that_caused_error = received_err
        .nlmsg
        .get_payload_as::<Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>>();
    let nlmsg_that_caused_error = nlmsg_that_caused_error.expect("Must contain original payload");
    let attr_handle = nlmsg_that_caused_error.get_attr_handle();
    println!(
        "ECHO-attribute was: {}",
        attr_handle
            .get_attr_payload_as::<String>(NlFoobarXmplAttribute::Msg)
            .expect("This attribute was sent and must be in the response")
    );
```

Currently there is a little divergence between the "ok" part of the API and the "err" part of the API. In the "ok" part, the type is e.g. `Nlmsghdr<u16, Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>>` whereas the type in the "err" part is `NlmsghdrErr` where you can get `Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>` from. Is this ideal? But I totally understand that your solution as it is right now is easy to integrate into the current code base without breaking changes.

Maybe it's part of a higher-level API to cope with this (request module for example)?

**TL;DR**: Looks good to me and it works! No more errors and I can receive multiple arbitrary receive messages, e.g. "ok" and "err", without `neli` errors.

jbaublitz:
> Thanks for working on this! It looks good now. FYI: This is the code I use to extract the error:
> 
> [phip1611/generic-netlink-user-kernel-rust@6adc643](https://github.com/phip1611/generic-netlink-user-kernel-rust/commit/6adc6430a60dbb280441566c72178b4b3cf440c2)
> 
> Or just the relevant snippet:
> 
> ```rust
>     let res: Result<
>         Option<Nlmsghdr<u16, Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>>>,
>         NlError,
>     > = sock.recv();
>     let received_err = res.unwrap_err();
>     let received_err = match received_err {
>         NlError::Nlmsgerr(err) => err,
>         _ => panic!("We expected an error here!"),
>     };
>     println!(
>         "Inside the kernel this error code occurred: {}",
>         received_err.error
>     );
>     let nlmsg_that_caused_error = received_err
>         .nlmsg
>         .get_payload_as::<Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>>();
>     let nlmsg_that_caused_error = nlmsg_that_caused_error.expect("Must contain original payload");
>     let attr_handle = nlmsg_that_caused_error.get_attr_handle();
>     println!(
>         "ECHO-attribute was: {}",
>         attr_handle
>             .get_attr_payload_as::<String>(NlFoobarXmplAttribute::Msg)
>             .expect("This attribute was sent and must be in the response")
>     );
> ```
> 
> Currently there is a little divergence between the "ok" part of the API and the "err" part of the API. In the "ok" part, the type is e.g. `Nlmsghdr<u16, Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>>` whereas the type in the "err" part is `NlmsghdrErr` where you can get `Genlmsghdr<NlFoobarXmplCommand, NlFoobarXmplAttribute>` from. Is this ideal? But I totally understand that your solution as it is right now is easy to integrate into the current code base without breaking changes.

I think the problem here is that it's impossible to type the error message appropriately for a couple reasons:
1. The length `nl_len` returned in the `NlmsghdrErr` structure is incorrect for the ACK case given that the payload is not actually returned for the message that triggered the ACK and we need to support both ACKs and errors.
2. It would be impossible to specify the appropriate type for all `NlError::Nlmsgerr` instances across the code base given that type information is erased at compile time so there's no way to preserve a type parameter between the `.send()` and `.recv()` calls.

> 
> Maybe it's part of a higher-level API to cope with this (request module for example)?

That was my intention. :) This is entirely possible to do there and I intend to do that.

> 
> **TL;DR**: Looks good to me and it works! No more `NlError` when I receive an error response with a Generic Netlink payload and I can receive multiple arbitrary messages in a row, e.g. "ok" and "err", without `unparsed data left in buffer` errors.

I'm so glad to hear this fixed it for you! It took me a bit to come up with a solution so I'm glad this helps! I'll get the documentation updated today and merge this.

phip1611:
@jbaublitz Do you think we should provide an example in the repo about this? An example that contains the snippet from above, with possibly a few adjustments to fit your desired coding conventions.

jbaublitz:
@phip1611 I've just pushed the additional documentation and example to this branch. If you're okay with this, I'll merge it.

phip1611:
@jbaublitz looks fine to me 👍

PR #148, opened 2021-04-01
route-list example: add interface name

MaxVerevkin:
## Before
```
default via 192.168.0.1 
192.168.0.0/24  proto Kernel  scope Link  src 192.168.0.106 
```

## After
```
default via 192.168.0.1 dev wlp3s0 
192.168.0.0/24 dev wlp3s0 proto Kernel scope Link src 192.168.0.106
```

## `ip route` for reference
```
default via 192.168.0.1 dev wlp3s0 proto dhcp metric 600 
192.168.0.0/24 dev wlp3s0 proto kernel scope link src 192.168.0.106 metric 600 
```

### DIFF
--- a/examples/route-list.rs
+++ b/examples/route-list.rs
@@ -1,6 +1,6 @@
 extern crate neli;
 
-use std::{error::Error, net::IpAddr};
+use std::{convert::TryInto, error::Error, net::IpAddr};
 
 use neli::{
     consts::{nl::*, rtnl::*, socket::*},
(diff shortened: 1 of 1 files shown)


jbaublitz (changes requested):
My understanding is that the interface name is passed back as a separate attribute. I don't think there's any need for unsafe code here. I'll double check on this though.

MaxVerevkin:
@jbaublitz Did you find a better way to get interface name?

jbaublitz:
Hi @MaxVerevkin, I'm so sorry for the delay. I've been away from this project for a while working on the v0.6.0 release. Please see #157 for the solution purely using netlink.

PR #150, opened 2021-04-04
getips: Don't ignore errors returned after iteration starts

joshtriplett:
getips ignores anything that isn't an NlPayload::Payload, which includes
NlPayload::Err. Instead, skip over NlPayload::Empty but process other
errors as errors.


jbaublitz (approved):
This looks like an improvement! Thanks so much!

Issue #152, opened 2021-08-07
Nlmgs containing Tcmsg fails serialization


tammela:
**Version of neli**
0.5.3

**Does the documentation specify that neil's behavior is correct?**
Not related.

**Does there appear to be a mistake in the netlink documentation based on the kernel code?**
Not related.

**Describe the bug**
Serialization of Tcmsgs seems to be bugged.
It's reporting `The number of bytes written to the buffer did not fill the given space` for very simple messages.

In an example, I have noticed that the buffer is expecting 40 bytes to be written as it's aligned to 4 bytes.
The message however is reporting 39 bytes, perhaps it's missing some padding?

**To Reproduce**
```
        let msg = Tcmsg {
            tcm_family: 0,
            tcm_ifindex: 0,
            tcm_handle: 0,
            tcm_parent: 0,
            tcm_info: 0,
            rtattrs: RtBuffer::new(),
            padding_char: 0,
            padding_short: 0,
        };
        
    let mut socket = NlSocketHandle::connect(socket::NlFamily::Route, None, &[])?;
    let nlhdr = {
        let nl_type = Rtm::Newtfilter;
        let nl_flags = nl::NlmFFlags::new(&[nl::NlmF::Request, nl::NlmF::Ack, nl::NlmF::Excl, nl::NlmF::Create]);
        let payload = NlPayload::Payload(msg);
        Nlmsghdr::new(None, nl_type, nl_flags, None, None, payload)
    };

    println!("{:?}", nlhdr);

    socket.send(nlhdr)?;
```

**Additional context**
I modified the library to expose the padding fields so I could build the Tcmsg payload.
See https://github.com/jbaublitz/neli/issues/151.

jbaublitz:
@tammela Can you please take a look and see if this is still a problem in v0.6.0? I've heavily improved the serialization code. If it hasn't, I'll fix it and add some tests.

jbaublitz:
Considering this closed by 16854b52d52fefed8e3ce3337b58499d23723934

Issue #151, opened 2021-08-07
Tcmsg is impossible to build in let statement


tammela:
**Version of neli**
0.5.3

**Does the documentation specify that neil's behavior is correct?**
No.

**Does there appear to be a mistake in the netlink documentation based on the kernel code?**
No.

**Describe the bug**
It's impossible to build Tcmsgs to send out in the socket due to  private padding fields.

**To Reproduce**
```
        let msg = Tcmsg {
            tcm_family: 0,
            tcm_ifindex: 0 as i32,
            tcm_handle: 0,
            tcm_parent: 0,
            tcm_info: 0,
            rtattrs: RtBuffer::new(),
            // Error!
        };
```

PR #153, opened 2021-08-08
Fixes for Tcmsg

tammela:
Fixes https://github.com/jbaublitz/neli/issues/152 and https://github.com/jbaublitz/neli/issues/151


jbaublitz:
Hi @tammela! Can you explain why you removed the padding bytes from the size?

PR #160, opened 2021-12-08
Implement FromBytes for &str and String

Kloenk:
This allows to retreive a String from an Attrhandle


Kloenk:
This could be simpler implemented with https://docs.rs/neli/latest/neli/attr/enum.AttrHandle.html#method.get_attr_payload_as_with_len . Will have to test it soon

jbaublitz:
I'm going to close this out due to the method you linked above.

PR #162, opened 2021-12-16
Fix initial size calculation for Rtattr

### LINKED ISSUE #158: RtBuffer encoded with invalid length
**Version of neli**
v0.6 & main

**Does the documentation specify that neil's behavior is correct?**
no, the kernel errors out with `netlink: 'linux::netlink:': attribute type 29 has an invalid length.` and Wireshark with `<MISSING>`

**Does there appear to be a mistake in the netlink documentation based on the kernel code?**
No

**Describe the bug**
RtAttributes are encoded with invalid length. Having one attribute with type `u32` reports length of 4, which is only the attribute header and the length. Correct value would be 8

**To Reproduce**
1. Request Getlink with `Ifla::ExtMask`
2. See dmesg

**Expected behavior**
Kernel should be able to parse the length.

jbaublitz:
Closes #158 
Closes #159 

@Kloenk Can you verify that this resolves #158? I made a PR to make the coding style more consistent with `Nlattr`.

### DIFF
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -274,12 +274,12 @@ where
     T: RtaType,
 {
     /// Create a new [`Rtattr`].
-    pub fn new<P>(rta_len: Option<u16>, rta_type: T, rta_payload: P) -> Result<Self, SerError>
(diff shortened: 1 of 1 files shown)


Kloenk (approved):
The tests in my project are succeeding. LGTM

PR #167, opened 2022-01-09
Fix bug in tokio usage and failure to handle empty netlink packet padding

### LINKED ISSUE #165: neli::tokio::NlSocket can't decode NLMSG_DONE.
**Version of neli**
v0.6.0

**Does the documentation specify that neil's behavior is correct?**
It's unclear.

> Netlink is a datagram-oriented service.  Both SOCK_RAW and SOCK_DGRAM are valid values for socket_type.  However, the netlink protocol does not distinguish between datagram and raw sockets.

But:
> Netlink messages consist of a byte stream with one or multiple nlmsghdr headers and associated payload.

**Does there appear to be a mistake in the netlink documentation based on the kernel code?**

I'm having trouble tracing exactly what's going on, so I'm not entirely sure who is at fault here, however parts of neli work and other parts don't, so I suspect the problem is somewhere in neli.

**Describe the bug**
`neli::socket::tokio::NlSocket` appears to fail with:
```
Wrapped(
    IOError(
        Error {
            kind: UnexpectedEof,
            message: "failed to fill whole buffer",
        },
    ),
)
```
when reading `NLMSG_DONE` messages.  Compared to `neli::socket::NlSocket` which seems fine.

I've not managed to figure out what is saying "failed to fill the whole buffer", I suspect the error message is coming from `std::io::Cursor` somewhere inside the macro expansion of `from_bytes_with_input`, but I'm unsure.

**To Reproduce**
Program fails, using neli::socket::tokio::NlSocket:
```rust
use neli::socket::tokio::NlSocket;
use neli::{
    consts::{
        nl::{GenlId, NlmF, NlmFFlags},
        socket::NlFamily,
    },
    err::DeError,
    genl::Genlmsghdr,
    nl::{NlPayload, Nlmsghdr},
    socket::NlSocketHandle,
    types::{GenlBuffer, NlBuffer},
};

#[neli::neli_enum(serialized_type = "u8")]
pub enum Nl80211Command {
    Unspecified = 0,
    GetWiPhy = 1,
    /* Many many more elided */
}
impl neli::consts::genl::Cmd for Nl80211Command {}

#[neli::neli_enum(serialized_type = "u16")]
pub enum Nl80211Attribute {
    Unspecified = 0,

    Wiphy = 1,
    /* Literally hundreds elided */
}
impl neli::consts::genl::NlAttrType for Nl80211Attribute {}

fn handle(msg: Nlmsghdr<GenlId, Genlmsghdr<Nl80211Command, Nl80211Attribute>>) {
    println!("msg={:?}", msg.nl_type);
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut sock = NlSocketHandle::connect(
        NlFamily::Generic, /* family */
        Some(0),           /* pid */
        &[],               /* groups */
    )?;
    let family_id = sock.resolve_genl_family("nl80211")?;

    let mut ss = NlSocket::new(sock)?;

    let req = &Nlmsghdr::new(
        /* len */ None,
        /* type */ family_id,
        /* flags */ NlmFFlags::new(&[NlmF::Request, NlmF::Dump, NlmF::Ack]),
        /* seq */ Some(1),
        /* pid */ Some(0),
        /* payload */
        NlPayload::Payload(Genlmsghdr::<Nl80211Command, Nl80211Attribute>::new(
            /* cmd */ Nl80211Command::GetWiPhy,
            /* version */ 1,
            /* attrs */ GenlBuffer::new(),
        )),
    );

    ss.send(req).await?;

    let mut buffer = Vec::new();

    loop {
        let data = ss.recv(&mut buffer).await;
        match data {
            Ok(msgs) => {
                println!("msgs: {:?}", msgs);
                for msg in msgs {
                    handle(msg);
                }
            }
            Err(e) => {
                println!("err: {:#?}", e);
            }
        }
    }
}
```

Compare/contrast the same program using `neli::socket::NlSocketHandle`:
```rust
use neli::socket::NlSocket;
use neli::{
    consts::{
        nl::{GenlId, NlmF, NlmFFlags},
        socket::NlFamily,
    },
    err::DeError,
    genl::Genlmsghdr,
    nl::{NlPayload, Nlmsghdr},
    socket::NlSocketHandle,
    types::{GenlBuffer, NlBuffer},
};

#[neli::neli_enum(serialized_type = "u8")]
pub enum Nl80211Command {
    Unspecified = 0,
    GetWiPhy = 1,
    /* Many many more elided */
}
impl neli::consts::genl::Cmd for Nl80211Command {}

#[neli::neli_enum(serialized_type = "u16")]
pub enum Nl80211Attribute {
    Unspecified = 0,

    Wiphy = 1,
    /* Literally hundreds elided */
}
impl neli::consts::genl::NlAttrType for Nl80211Attribute {}

fn handle(msg: Nlmsghdr<GenlId, Genlmsghdr<Nl80211Command, Nl80211Attribute>>) {
    println!("msg={:?}", msg.nl_type);
}

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut sock = NlSocketHandle::connect(
        NlFamily::Generic, /* family */
        Some(0),           /* pid */
        &[],               /* groups */
    )?;
    let family_id = sock.resolve_genl_family("nl80211")?;

    let req = Nlmsghdr::new(
        /* len */ None,
        /* type */ family_id,
        /* flags */ NlmFFlags::new(&[NlmF::Request, NlmF::Dump, NlmF::Ack]),
        /* seq */ Some(1),
        /* pid */ Some(0),
        /* payload */
        NlPayload::Payload(Genlmsghdr::<Nl80211Command, Nl80211Attribute>::new(
            /* cmd */ Nl80211Command::GetWiPhy,
            /* version */ 1,
            /* attrs */ GenlBuffer::new(),
        )),
   );

    sock.send(req)?;

    loop {
        let data = sock.recv();
        match data {
            Ok(msgs) => {
                println!("msgs: {:?}", msgs);
                for msg in msgs {
                    handle(msg);
                }
            }
            Err(e) => {
                println!("err: {:#?}", e);
            }
        }
    }
}
```

This demonstrates the problem both on machines without wifi interfaces, and machines with wifi interfaces (since the NLMSG_DONE message appears in both).  It should not require root or special privileges to run.

**Expected behavior**
Both programs should end with NLMSG_DONE (which these programs show as UnrecognizedConst(3)).  Instead only the non-async version does so, and the async version ends with the error "failed to fill the whole buffer".

### LINKED ISSUE #166: poll_read_priv unconditionally clear_ready!()
**Version of neli**
0.6.0

**Does the documentation specify that neil's behavior is correct?**
No - tokio says you only should clear the ready signal when the socket would block.

**Does there appear to be a mistake in the netlink documentation based on the kernel code?**
This is independent of netlink, it's a misuse of the tokio apis.

**Describe the bug**

neli::socket::tokio::NlSocket uses an AsyncFd to manage the events on the socket:

```rust
    fn poll_read_priv(
        async_fd: &AsyncFd<super::NlSocket>,
        cx: &mut Context,
        buf: &mut ReadBuf,
    ) -> Poll<io::Result<usize>> {
        let mut guard = ready!(async_fd.poll_read_ready(cx))?;
        guard.clear_ready();
        let socket = async_fd.get_ref();
        let bytes_read = socket.recv(buf.initialized_mut(), 0)?;
        buf.advance(bytes_read);
        Poll::Ready(Ok(bytes_read))
    }
```

This calls `let mut guard = ready!(async_fd.poll_read_ready(cx))?;`, but then unconditionally calls `guard.clear_ready();`.  The AsyncFd documentation for clear_ready() says:

> It is critical that this function not be called unless your code actually observes that the file descriptor is not ready. Do not call it simply because, for example, a read succeeded; it should be called when a read is observed to block.

Instead this should call `socket.recv()` and inspect to see if the error return is `WouldBlock` and only call `clear_ready()` then.  I think this can be better done with try_io.

eg (untested code):
```rust
    fn poll_read_priv(
        async_fd: &AsyncFd<super::NlSocket>,
        cx: &mut Context,
        buf: &mut ReadBuf,
    ) -> Poll<io::Result<usize>> {
     loop {
        let mut guard = ready!(async_fd.poll_read_ready(cx))?;
        match guard.try_io(|socket| socket.get_ref().recv(buf.initialized_mut(), 0)) {
          Ok(Ok(bytes)) => { buf.advance(bytes); return Poll::Ready(Ok(bytes)); }
          Ok(result) => return Poll::Ready(result),
          Err(_would_block) => continue,
        }
      }
  }
```

Without this, if there are more bytes than the buffer has waiting, then the program blocks waiting for the fd to become ready again, but since it's already ready, epoll() never returns the read event to tokio, blocking the program.

The write version of the function seems to have the same bug (although I don't think I've ever triggered it), with basically the same fix.

**To Reproduce**
1. Send a request onto a netlink socket that has a reply, do not wait for the reply.
2. Immediately send another request onto the netlink socket that also has a reply.
3. Attempt to wait for both replies.  Note that only the first reply arrives, and the second does not.

**Expected behavior**
neli should not block forever waiting for a ready signal that will never come.

jbaublitz:
Closes #166
Closes #165 

@isomer Can you confirm that this fixes both bugs on your end?

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -4,7 +4,6 @@ cargo fmt -- --check \
     && cargo clippy --all-targets --all-features -- -D warnings \
     && cargo build \
     && cargo build --all-targets --all-features \
-    && cargo test \
     && cargo test --all-targets --all-features \
     || exit 1
 
@@ -16,7 +15,8 @@ cargo fmt -- --check \
     && ./target/debug/route-list \
     && ./target/debug/ctrl-list \
     && ./target/debug/error_packet \
-    && cargo build --all-targets --all-features) \
+    && cargo build --all-targets --all-features \
+    && ./target/debug/async_nl80211) \
     || exit 1
 
 (cd neli-proc-macros \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         task:
-          - cargo clippy --all-targets -- -D warnings
+          - cargo clippy --all-targets --all-features -- -D warnings
           - (cd neli-proc-macros && cargo clippy --all-targets -- -D warnings)
         toolchain:
           - nightly
@@ -38,8 +38,6 @@ jobs:
           - cargo build
           - cargo build --features=async
           - cargo build --all-features
-          - RUST_TEST_THREADS=1 cargo test
-          - RUST_TEST_THREADS=1 cargo test --features=async
           - RUST_TEST_THREADS=1 cargo test --all-features
           - (cd neli-proc-macros && cargo fmt -- --check)
           - (cd neli-proc-macros && cargo build)
@@ -88,8 +86,6 @@ jobs:
           - cargo build
           - cargo build --features=async
           - cargo build --all-features
-          - RUST_TEST_THREADS=1 cargo test
-          - RUST_TEST_THREADS=1 cargo test --features=async
           - RUST_TEST_THREADS=1 cargo test --all-features
         toolchain:
           - nightly
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,10 @@ optional = true
 simple_logger = "1.13.0"
 lazy_static = "1.4.0"
 
+[dev-dependencies.tokio]
+version = "1"
+features = ["macros", "rt-multi-thread"]
+
 [features]
 default = []
 async = ["tokio", "futures"]
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -36,9 +36,15 @@ path = "lookup_id.rs"
 name = "error_packet"
 path = "error_packet.rs"
 
+[[bin]]
+name = "async_nl80211"
+path = "async_nl80211.rs"
+required-features = ["async"]
+
 [dependencies.tokio]
-version = "1.0"
+version = "1"
 optional = true
+features = ["macros"]
 
 [dependencies.futures]
 version = "0.3"
--- a/examples/async_nl80211.rs
+++ b/examples/async_nl80211.rs
@@ -1,0 +1,89 @@
+use std::error::Error;
+
+#[cfg(feature = "logging")]
+use log::Level;
+#[cfg(feature = "logging")]
+use simple_logger::init_with_level;
+
+use neli::{
+    consts::{
+        nl::{GenlId, NlmF, NlmFFlags},
+        socket::NlFamily,
+    },
+    genl::Genlmsghdr,
+    nl::{NlPayload, Nlmsghdr},
+    socket::tokio::NlSocket,
+    socket::NlSocketHandle,
+    types::GenlBuffer,
+};
+
+#[neli::neli_enum(serialized_type = "u8")]
+pub enum Nl80211Command {
+    Unspecified = 0,
+    GetWiPhy = 1,
+    /* Many many more elided */
+}
+impl neli::consts::genl::Cmd for Nl80211Command {}
+
+#[neli::neli_enum(serialized_type = "u16")]
+pub enum Nl80211Attribute {
+    Unspecified = 0,
+
+    Wiphy = 1,
+    /* Literally hundreds elided */
+}
+impl neli::consts::genl::NlAttrType for Nl80211Attribute {}
+
+fn handle(msg: Nlmsghdr<GenlId, Genlmsghdr<Nl80211Command, Nl80211Attribute>>) {
+    println!("msg={:?}", msg.nl_type);
+}
+
+#[tokio::main]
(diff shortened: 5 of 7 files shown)


isomer:
Yup!  This seems to work well, thank you very much for the quick fixes!

jbaublitz:
@isomer I think I will have to disable the test in CI for now due to nl80211 not being available on Github Actions infrastructure, but I'll include it in the git hooks so that if anything breaks, I will catch it when I go to commit.

isomer:
yeah, it appears to return ENOENT when you don't have any 802.11 support, and returns an empty list if you support 802.11 but have no drivers.  And I can kinda see why CI images might decide that they're unlikely to discover an 802.11 driver/network.

PR #164, opened 2022-01-01
Fix bug of GenlBuffer Size implementation

reticenceji:
Hi, I found that there is a bug in computation of `nlmsg_len` of a netlink package.

If we use generic netlink, the payload is in format of Type-Length-Value attribute. 
We must align the attributes to 4 bytes.
So the `nlmsg_len` should sum up all **aligned attributes**' size, but not sum up all attributes' size then align.

English is not my native language, and I apologize for any grammatical errors and improper wording.

Here is an example can produce the bug.
```rust
mod test {
    use neli::{
        consts::nl::{NlmF, NlmFFlags},
        genl::{Genlmsghdr, Nlattr},
        neli_enum,
        nl::{NlPayload, Nlmsghdr},
        types::{Buffer, GenlBuffer},
    };

    #[neli_enum(serialized_type = "u16")]
    enum SendAttribute {
        Unspec = 0,
        A = 1,
        B = 2,
    }
    impl neli::consts::genl::NlAttrType for SendAttribute {}

    #[test]
    fn test() {
        let mut attrs: GenlBuffer<SendAttribute, Buffer> = GenlBuffer::new();
        let attr_a = "123"; // 4 bytes
        let attr_b = "123";

        attrs.push(Nlattr::new(false, false, SendAttribute::A, attr_a).unwrap());
        attrs.push(Nlattr::new(false, false, SendAttribute::B, attr_b).unwrap());
        let gnmsghdr = Genlmsghdr::new(1, 1, attrs);

        let nlmsghdr = Nlmsghdr::new(
            None,
            1,
            NlmFFlags::new(&[NlmF::Request]),
            Some(0),
            Some(1),
            NlPayload::Payload(gnmsghdr),
        );
        assert_eq!(nlmsghdr.nl_len,36);

        let mut attrs: GenlBuffer<SendAttribute, Buffer> = GenlBuffer::new();
        let attr_a = "1234"; // 5 bytes, should take 8 bytes
        let attr_b = "123";

        attrs.push(Nlattr::new(false, false, SendAttribute::A, attr_a).unwrap());
        attrs.push(Nlattr::new(false, false, SendAttribute::B, attr_b).unwrap());
        let gnmsghdr = Genlmsghdr::new(1, 1, attrs);

        let nlmsghdr = Nlmsghdr::new(
            None,
            1,
            NlmFFlags::new(&[NlmF::Request]),
            Some(0),
            Some(1),
            NlPayload::Payload(gnmsghdr),
        );
[...]

PR #169, opened 2022-01-16
Fix alignment issues in attribute size calculation

jbaublitz:
Closes #164

@reticenceji Can you take a look at this PR? I've kept your commits to give you credit for reporting and fixing the bug and added an additional commit to clean up some of the usage of size calculation. If this resolves your problem, I'll merge it!


reticenceji:
It solved my problem! :)

PR #175, opened 2022-03-05
remove futures dependency

MaxVerevkin:


jbaublitz:
Hi @MaxVerevkin! Could you explain this PR a little bit more? What's your rationale for these changes?

MaxVerevkin:
1) I just removed the unused tokio features because why should we keep them around?
2) I removed the futures dependency because the only thing we need from it is the ready macro, which is short and simple enough to just copy-paste. This should result in a bit better compile times, I suppose. I didn't benchmark this.

jbaublitz (approved):
Ok, this looks good to me! Thanks for the explanation.

PR #177, opened 2022-03-15
Add documentation for NlmF::Request and fix neli-proc-macros to support this

### LINKED ISSUE #133: Always add NlmF::Request as default behaviour (or if no flags are present) because Kernel requires it and libnl does it too
`Nlmsghdr::new()` takes field `nl_flags: NlmFFlags` which can be initialized like this: `NlmFFlags::new(&[NlmF::Request]),`. If the provided array is empty, i.e. `REQUEST (0x1)` is NOT present, the kernel rejects the request and doesn't forward it to the callback inside the kernel. Kernel reference: https://elixir.bootlin.com/linux/latest/source/net/netlink/af_netlink.c#L2487
This can result in a deadlock (socket/send never returns) in userland if the application waits for a reply in a blocking way.

`libnl`, the de facto standard library for netlink in C, also adds this flag automatically on each request: https://github.com/thom311/libnl/blob/2abeec851e4c748effac341f8db0b42389b37790/lib/nl.c#L489 

My suggestion:
- A: either add `NlmF::Request` always because there should be no usecase where the user doesn't want to send this flag, or
- B: make this an Option. If  `None` is provided it should be replaced with `NlmFFlags::new(&[NlmF::Request])`
- C: make this an Option and always add `NlmF::Request` as default flag (whether its None or Some) 

I prefer A or C because `neli` should help the user to make no mistakes at all.

PS: If you give me your ok, I can prepare this and create a PR.

jbaublitz:
Closes #133 

Hi @phip1611! Take a look if you'd like. I've added the documentation and it should resolve the above linked issue.

### DIFF
--- a/neli-proc-macros/src/neli_enum.rs
+++ b/neli-proc-macros/src/neli_enum.rs
@@ -5,6 +5,8 @@ use syn::{
     parse, parse_str, Arm, Attribute, Expr, Ident, ItemEnum, Lit, Meta, Path, Token, Type, Variant,
 };
 
+use crate::shared::remove_bad_attrs;
+
 fn parse_type_attr(attr: Meta) -> Type {
     if let Meta::NameValue(nv) = attr {
         if nv.path == parse_str::<Path>("serialized_type").unwrap() {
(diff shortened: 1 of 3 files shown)


phip1611:
@jbaublitz  looks good to me!

PR #182, opened 2022-06-25
v0.7.0 development

### LINKED ISSUE #83: Add convience methods and builder pattern for neli struct construction

### LINKED ISSUE #86: Make all neli struct fields private and generate/implement getters for fields

### LINKED ISSUE #107: Consider using bitflags crate in place of impl_flags macro and FlagBuffer
It doesn't seem necessary to store flags in a Vec and bitwise-or them together when serializing. Flags could be stored as bitflags all the time.

It might be possible to switch to the bitflags crate rather than reimplementing a bitflags mechanism. That would also provide additional convenience methods, such as converting between the bitflags wrapper type and the underlying numeric type.

Enough of the implementation details of `impl_flags!` are hidden that it *might* be possible to do this without breaking backwards compatibility.

### LINKED ISSUE #124: Remove nested argument for NlAttr::new
NlAttr::new accepts an argument to set the "has nested attributes" flag for the attribute. However, this flag should generally match the presence of nested attributes. NlAttr::new could drop that argument, and instead automatically set the flag to `true` when adding a nested attribute.

For the benefit of kernel protocols that handle the nested flag incorrectly, there could be a `set_nested` method that directly sets the flag to true or false, ignoring whether there are actually nested attributes. This method's documentation should explicitly state that it doesn't normally need to be called, that the nested flag is set automatically, and that it should only be called to work around kernel bugs. (It would be nice if that documentation could make reference to at least one *specific* kernel netlink protocol that's broken in that way.)

### LINKED ISSUE #161: rta_len is a dead parameter
**Version of neli**
0.6.x

**Describe the bug**
This parameter should be removed in 0.7.0.

Related to #158

### LINKED ISSUE #168: Padded size procedural macro feature
No

#164 

There need to be some changes to the `Size` procedural macro to allow defining a field as padded when the size is calculated. This is particularly necessary for situations where the container type expects the internal data structure to be padded so it must always be calculated that way.

### LINKED ISSUE #170: Support ExtAcks
Maybe - Depends on how it's implemented.

I'm always frustrated when I get errors like "EINVAL" back from the kernel.  Figuring out exactly what is invalid is complicated and annoying.

The kernel appears to support "extacks" which are extended error strings on netlink messages: https://lwn.net/Articles/719721/ / https://grep.app/search?q=NL_SET_ERR_MSG

This is enabled with a setsockopt:
```
  setsockopt(fd, SOL_NETLINK, NETLINK_EXT_ACK, true);
```

Then extra attributes are included on the netlink error messages:
```c
struct nlmsgerr {
        int             error;
        struct nlmsghdr msg;
        /*
         * followed by the message contents unless NETLINK_CAP_ACK was set
         * or the ACK indicates success (error == 0)
         * message length is aligned with NLMSG_ALIGN()
         */
        /*
         * followed by TLVs defined in enum nlmsgerr_attrs
         * if NETLINK_EXT_ACK was set
         */
};

/**
 * enum nlmsgerr_attrs - nlmsgerr attributes
 * @NLMSGERR_ATTR_UNUSED: unused
 * @NLMSGERR_ATTR_MSG: error message string (string)
 * @NLMSGERR_ATTR_OFFS: offset of the invalid attribute in the original
 *       message, counting from the beginning of the header (u32)
 * @NLMSGERR_ATTR_COOKIE: arbitrary subsystem specific cookie to
 *      be used - in the success case - to identify a created
 *      object or operation or similar (binary)
 * @NLMSGERR_ATTR_POLICY: policy for a rejected attribute
 * @__NLMSGERR_ATTR_MAX: number of attributes
 * @NLMSGERR_ATTR_MAX: highest attribute number
 */
```
  -- /usr/include/linux/netlink.h

Instead of NlPayload::Err() returning an std::io::Error, it could return a different Error type that includes all the extra metadata, and a nested std::io::Error.

This would drastically help with providing better error messages, and would avoid me pulling my hair out when I screw up and get yet another mysterious EINVAL.

The obvious alternative would not be to implement ExtAck support.

But more usefully, another option would be for NlPayload::Err to continue to implement a std::io::Error but the msg would be from NLERRMSG_ATTR_MSG instead of "Invalid Argument" (although I can't see a way in the standard library to do that).

Another variant would be to have std::io::Error wrap a different error that contains more information but continues to set ErrorKind correctly.

### LINKED ISSUE #194: Proper way of `addattr32`
**Is this a question about neli specifically? If so, please describe.**

I'm trying to do a RTM_GETNSID query with neli, and have problem adding a NETNSA_FD attribute after Rtgenmsg. I managed to bypass the problem by adding a Nlattr manually, with the following code:

```
    let mut c = Cursor::new(Vec::new());
    let _ = Rtgenmsg {
        rtgen_family: RtAddrFamily::Unspecified,
    }.to_bytes(&mut c);
    c.write(&[0, 0, 0]);

    let file = File::open(path).unwrap();
    Nlattr::new(false, false, 3, file.as_raw_fd()).unwrap().to_bytes(&mut c);

    let buf = c.into_inner();
    let req = Nlmsghdr::new(
        None,
        Rtm::Getnsid,
        NlmFFlags::new(&[NlmF::Request]),
        None,
        None,
        NlPayload::Payload(buf.as_slice()),
    );
```

The equivalent code in C looks like:
```
	struct {
		struct nlmsghdr n;
		struct rtgenmsg g;
		char            buf[1024];
	} req = {
		.n.nlmsg_len = NLMSG_LENGTH(sizeof(struct rtgenmsg)),
		.n.nlmsg_flags = NLM_F_REQUEST,
		.n.nlmsg_type = RTM_GETNSID,
		.g.rtgen_family = AF_UNSPEC,
	};
	int fd = -1;

	fd = netns_get_fd(name);
	if (fd < 0)
		return fd;

	addattr32(&req.n, 1024, NETNSA_FD, fd);
```

Sorry if this is a dumb question.

### LINKED ISSUE #195: Parsing different NlAttrType Enums for different received Cmds
Hey,

I'm sorry if the response to this is somewhere in the docs, but i searched them and read a lot of issues but still could not find a solution for my problem.
Issue #50 seems to be relevant, but could not find an exact solution after reading it.

Neli version: 0.6.2

I am using the async Socket neli::socket::tokio::NlSocket to wait for messages from the socket via:
```
let mut buffer: Vec<u8> = Vec::new();
let response =  self.socket.recv::<NlTypeWrapper, Genlmsghdr<MYCMDS, ATTRIBUTES_COMMAND_ONE>>(&mut buffer);
```

Now i am forced to give socket the type of the attributes. But i would like to parse the attributes of the Genlmsghdr depending on the command i got in MYCMDS.

Right now my attribute enums and commands look something like this:
```
#[neli_enum(serialized_type = "u8")]
pub enum MYCMDS {
    COMMAND_ONE = 0,
    COMMAND_TWO = 1,
}
impl Cmd for MYCMDS {}

#[neli_enum(serialized_type = "u16")]
pub enum ATTRIBUTES_COMMAND_ONE {
    ATTRIBUTE_ONE_A = 0,
    ATTRIBUTE_ONE_B = 1,
}
impl NlAttrType for ATTRIBUTES_COMMAND_ONE {}

#[neli_enum(serialized_type = "u16")]
pub enum ATTRIBUTES_COMMAND_TWO {
    ATTRIBUTE_TWO_A = 0,
    ATTRIBUTE_TWO_B = 1,
    ATTRIBUTE_TWO_C = 2,
    ATTRIBUTE_TWO_D = 3,
}
impl NlAttrType for ATTRIBUTES_COMMAND_TWO {}
```

And i would like to do something like this (maybe not correct code. just as an example):
```
let response =  self.socket.recv::<NlTypeWrapper, Genlmsghdr<MYCMDS, u16>>(&mut buffer);
let message = response.unwrap()
let attr_handle = match nlmessage.nl_payload.get_payload().unwrap().cmd{
    MYCMDS::COMMAND_ONE => {
        nlmessage.get_payload().unwrap().get_attr_handle_as(ATTRIBUTES_COMMAND_ONE)
    }
    MYCMDS::COMMAND_TWO => {
        nlmessage.get_payload().unwrap().get_attr_handle_as(ATTRIBUTES_COMMAND_TWO)
    }
  }

```

Hope you understand my use case.
Thanks in advance!

jbaublitz:
Closes #107
Closes #170 
Closes #124 
Closes #161
Closes #83 
Closes #168 
Closes #194 
Closes #86 
Closes #195


src/err.rs
```diff
@@ -63,15 +63,18 @@ pub struct NlmsghdrErr<T, P> {
...
 pub struct Nlmsgerr<T, P> {
     /// Error code
     pub error: libc::c_int,
     /// Packet header for request that failed
-    #[neli(input = "input - Self::header_size()")]
+    #[neli(input = "input - std::mem::size_of::<libc::c_int>()")]
     pub nlmsg: NlmsghdrErr<T, P>,
+    #[neli(input = "input - std::mem::size_of::<libc::c_int>() - nlmsg.unpadded_size()")]
+    /// Contains attributes representing the extended ACK
+    pub ext_ack: GenlBuffer<u16, Buffer>,
```
isomer, line 77: Should this be:
```rust
    pub ext_ack: GenlBuffer<NlmsgerrAttr, Buffer>,
```

with somewhere in consts:
```rust
// from enum nlmsgerr_attrs in <linux/netlink.h>
#[neli_enum(serialized_type = "u16")]
enum NlmsgerrAttr {
         Unused = 0,
         /// error message string (string)
         Msg = 1,
         /// offset of the invalid attribute in the original
         /// message, counting from the beginning of the
         /// header (u32)
         Offset = 2,
         /// arbitrary subsystem specific cookie to
         /// be used - in the success case - to identify a created
         /// object or operation or similar (binary)
         Cookie = 3,
         /// policy for a rejected attribute
         Policy = 4,
}
```

jbaublitz: Yes that is what I want the interface to look like. I mainly just wanted a prototype to see if it works for you first. It seems like it does, so I'll add this in!

jbaublitz: Resolved!

PR #186, opened 2022-07-12
Allow converting flag sets from bitmasks

### LINKED ISSUE #178: Add from<libc::c_uint> for IffFlags
No.

`rtnetlink(7)` describes Ifinfomsg.ifi_change as:
```
ifi_change is reserved for future use and should be always set to 0xFFFFFFFF.
```

However this is ugly to do with the existing IffFlags type, the best solution I've found is:
```
IffFlags::new(&[Iff::UnrecognizedConst(0xFFFFFFFF)]),
```
This is setting a flag with multiple bits, which seems wrong.

Implement `From<libc::c_uint> for IffFlags` (via implementing `From<ty>` in `impl_flags!`), so I instead can write:
```
 0xFFFFFFFF.into(),
```
when constructing `Ifinfomsg`.

Ideally, there would be a `From<IffFlags> for libc::c_uint` too (for symmetry) but this isn't necessary.

 * have `IffFlags::new()` take an `Option<>` for `ifi_changed`, and set it to 0xFFFFFFFF internally if not specified.  This is a breaking change, and not particularly general.  A variation on this, since ifi_changed is not yet used, to just drop the parameter from `IffFlags::new()` and hard code the value to 0xFFFFFFFF.
 *  have impl_flags! create a `full()` method, to create symmetry with the existing `empty()`.  This isn't as general as just allowing you to convert back and forth to the underlying type, and seems like pointless symmetry.

jbaublitz:
Closes #178 

@isomer Any thoughts?

### DIFF
--- a/src/consts/macros.rs
+++ b/src/consts/macros.rs
@@ -220,6 +220,11 @@ macro_rules! impl_flags {
                 $name($crate::types::FlagBuffer::<$bin_type, $type>::empty())
             }
 
+            /// Create a flag container from a bitmask.
+            pub fn from_bitmask(bitmask: $bin_type) -> Self {
+                $name($crate::types::FlagBuffer::<$bin_type, $type>::from_bitmask(bitmask))
+            }
+
             /// Initialize a flag container with the given flags
             pub fn new(flags: &[$type]) -> Self {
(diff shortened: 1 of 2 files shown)


isomer:
Sorry for the delay in replying.

This looks great, thank you very much!

PR #200, opened 2022-10-26
Fix error field interpretation

xpromache:
According to man 7 netlink, in case of error, the error field represents "negative errno". Also a simple test trying to add an interface as an unprivileged user, I get that field as -1 which the Display function translates into "Unknown error". With the change proposed, I get  "Operation not permitted" which is what one would expect.


jbaublitz:
I feel like I fixed this a while back, so I'm wondering if this is a regression. Thanks for putting this change up and I'll release a patch release shortly!

PR #213, opened 2023-02-26
Add methods to control strict input checking for route sockets

cgzones:
Supported since Linux 4.20.

See https://kernel.org/doc/html/next/userspace-api/netlink/intro.html#strict-checking


jbaublitz:
Hi @cgzones, thanks for the contribution! If I'm understanding this correctly, this basically is a toggle to validate all attributes for routing netlink even if they are not needed by the kernel for the given request, is that correct?

cgzones:
To the best of my knowledge, yes.

jbaublitz (approved):

PR #219, opened 2023-07-01
send all multicast events

### LINKED ISSUE #218: Route multicast events (with `nl_pid > 0`) are ignored
I'm trying to make something that functions similar to [`nl-monitor ipv4-ifaddr`](https://github.com/thom311/libnl/blob/main/src/nl-monitor.c) with `neli`, but I'm struggling to get anywhere with it.

This is my current attempt:

```rust
let (socket, mut multicast) =
    NlRouter::connect(NlFamily::Route, None, Groups::empty()).unwrap();

socket
    .add_mcast_membership(Groups::new_groups(&[RTNLGRP_IPV4_IFADDR]))
    .unwrap();

match multicast.next() {
    None => unreachable!(),
    Some(response) => {
        dbg!(response).unwrap();
    }
}
```

As far as I can tell, this should be reporting events any time an ipv4 address on the machine changes, but I get no output at all with this setup. I've been looking at how `nl-monitor` works, and comparing that to `neli` and they look very similar here, so I'm not sure what's different...

Any chance you might know where I should look next? :pray:

acheronfail:
This fixes #218 

Let me know if you want this to be a configurable option, or if you want any other changes.

### DIFF
--- a/src/router/asynchronous.rs
+++ b/src/router/asynchronous.rs
@@ -64,11 +64,12 @@ fn spawn_processing_thread(socket: Arc<NlSocketHandle>, senders: Senders) -> Pro
                             Ok(m) => {
                                 let seq = *m.nl_seq();
                                 let lock = senders.lock().await;
-                                if m.nl_pid() == &0 {
+                                let seq_sender = lock.get(m.nl_seq());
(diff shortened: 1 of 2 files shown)


acheronfail:
The actual solution is in #209 !

PR #220, opened 2023-07-07
`get_nested_attributes` doesn't need `&mut self`

acheronfail:
While using this I noticed that the `.get_nested_attributes` took mutable references, but I don't think the `mut` is required here. 

This PR removes it (and _**targets the `v0.7.0-rc2` branch**_).

---

By the way, `v0.7.0-rc2` is looking really good IMO!


jbaublitz:
@acheronfail Thanks for catching this! Could you also apply the change to `rtnl.rs`? The same problem exists there too.

acheronfail:
@jbaublitz unless I've missed another one, this PR already contains an update to `rtnl.rs` which removes the `mut`?

jbaublitz:
Whoops! I totally missed that. Merging now.

PR #234, opened 2023-11-03
Fix feature flags

### LINKED ISSUE #232: 0.7.0-rc2 does not build with `async` disabled
Trying to only enable `async` support, this fails to build wth numerous errors:
```
cargo build --no-default-features -F async
```

jbaublitz:
Closes #232


jbaublitz:
@ydirson This is now incorporated into CI so it should be tested moving forward.

PR #239, opened 2024-06-21
Migrate to syn 2.x

duMagnus:
The transition to syn 2.x involved addressing several breaking changes from syn 1.x. Some of these changes were straightforward, such as the replacement of Tokens Add with Plus, Colon2 with PathSep, and the renaming of LifeTimeDef struct to LifeTimeParam. However, some changes required modifications to our structures. For instance, we now retrieve .meta from an Attribute instead of using .parse_meta(). Additionally, Paths now support .is_ident(), which simplifies their comparison.

The removal of NestedMeta necessitated a new approach to parsing nested elements from Paths using TokenTree tokens. These tokens can be Groups, Idents, Puncts, or Literals, enabling us to determine their purpose.


duMagnus:
Hi @jbaublitz, I still want to fix some clippy lints and improve the commit message.

PR #263, opened 2025-01-24
Improve usability for Getlink and fix unit tests

jbaublitz:


compenguy:
Confirmed, switching to the `issue-261` branch, and relying on the default value of Arphrd allows me to get results for the Getlink request.

jbaublitz:
Just a note that you can do this on the currently release too if you set ifi_type to `Arphrd::from(0)`. It does not just need to be on this branch.

Issue #270, opened 2025-02-19
Batch send support


LogicalOverflow:
**Would adding this feature cause a breaking change?**
No

**Is your feature request related to a problem? Please describe.**
It's impossible to send multiple messages as a single socket send call through a `NlRouter` (or `NlSocketHandle`), which is, e.g., needed to send transactions changing nftables. (They consist of a start message, a list of change commands, and a end message, all of which are complete nlmsgs but must (afaict) be sent in a single send over the socket)

**Describe the solution you'd like**
A send method on the `NlRouter` which takes multiple messages to send as a batch. Not sure how to deal with the receive handle and sequence numbers here.

**Describe alternatives you've considered**
Currently, you need access to the underlying `NlSocket` to do such batch sends. A simpler solution would be adding a batch send method to `NlSocketHandle` to allow using at least this level of abstraction. Though this would it still make it impossible to use `NlRouter` when also sending batch messages.

jbaublitz:
Hi @LogicalOverflow, thanks for opening an issue about this. With regard to the sequence numbers, are you wondering whether they should be incremented or all the same? I'm happy to take a look at #271, but I think it would preferable to find a way to support this at the router level as well if possible. Do you know of any documentation for nftables that I could take a look at? If not, I'll see what I can find in the kernel code.

LogicalOverflow:
> With regard to the sequence numbers, are you wondering whether they should be incremented or all the same?

Yes, both solutions don't feel to great: If we don't increment for each message in the batch, we have duplicate sequence numbers; if we do increment, we have to deal with that in the receive handle or return multiple receive handles.

> I'm happy to take a look at https://github.com/jbaublitz/neli/pull/271, but I think it would preferable to find a way to support this at the router level as well if possible.

I would like router-level support as well, just didn't have time to properly implement that. #271 is more of an intermediate solution, so you can at least do batch sends over sockets.

> Do you know of any documentation for nftables that I could take a look at? If not, I'll see what I can find in the kernel code.

nftables docs is, let's say, very minimal, esp. wrt. to the netlink interface; the nftables wiki has some helpful info on concepts here and there, but kernel code seems to be the best resource overall for the netlink interface. And libnfnl sources can be helpful as well.

jbaublitz:
> > With regard to the sequence numbers, are you wondering whether they should be incremented or all the same?
> 
[...]

PR #271, opened 2025-02-19
Add send_batch method to NlSocketHandles

LogicalOverflow:
This is the easy part of #270, adding a method that allows sending multiple `Nlmsghdr` in a single socket send call.

I am not sure if you want `&[Nlmsghdr]` for the interface, or if something else would be nicer.

### DIFF
--- a/src/socket/asynchronous.rs
+++ b/src/socket/asynchronous.rs
@@ -91,6 +91,30 @@ impl NlSocketHandle {
         Ok(())
     }
 
+    /// Send multiple messages on the socket asynchronously.
+    pub async fn send_batch<T, P>(&self, msgs: &[Nlmsghdr<T, P>]) -> Result<(), SocketError>
+    where
+        T: NlType,
+        P: Size + ToBytes,
+    {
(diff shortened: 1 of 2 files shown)


jbaublitz (changes requested):
Would you mind adding a test for each method? I'm trying to be a little more careful about making sure that new code is tested while I work my way back through untested code and add new tests. After that, this should be good to merge.

LogicalOverflow:
Sorry for the delay as well, I'll take a look if I can find any (batch-able) endpoints we can easily call for the test. Batch sends are mostly (exclusively?) used for writes and writing firewall rules in a library test probably isn't a good idea. 

I'm also have to see what permissions will be needed to run the tests like this - would be nice if the test work with normal user privileges

jbaublitz:
My guess is that any kind of list that you can find would be able to be run as a regular user.

jbaublitz:
@LogicalOverflow I'm going to be putting out the release soon of 0.7.0. We can always include this in 0.7.1 if desired.

PR #276, opened 2025-04-10
fix: log naked structs at trace!() level

tchernobog:
The problem with logging at debug level is that we have neli as part of a much bigger binary, and right now neli dumps very big unformatted structs in syslog.

We believe that neli should instead log these items at a lower level, such as trace!(). Debug-level log messages are still relevant for end users of the system, but trace messages are only for developers that know their way around on the system and thus are expected to know how to filter noise out (e.g. via environment variables to the logger).

### DIFF
--- a/src/iter.rs
+++ b/src/iter.rs
@@ -2,7 +2,7 @@
 
 use std::{io::Cursor, marker::PhantomData};
 
-use log::debug;
+use log::trace;
 
 use crate::{
     consts::nl::NlType, err::SocketError, nl::Nlmsghdr, FromBytes, FromBytesWithInput, Size,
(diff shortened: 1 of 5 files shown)


tchernobog:
@jbaublitz ping

jbaublitz:
@tchernobog Just to confirm, you're targeting version 0.7.0, correct?

tchernobog:
@jbaublitz yes, 0.7.0 is fine. I mostly needs this through [the local_ip_address crate](https://docs.rs/local-ip-address/latest/local_ip_address/), so I will also have to patch that one once this is released.

jbaublitz:
@tchernobog If you could fix CI and then squash it all into one commit, I'll merge this.

jbaublitz (approved):

PR #277, opened 2025-05-29
Feature/connector

### LINKED ISSUE #275: Connector protocol
**Is this a question about neli specifically? If so, please describe.**

Hello, I'm trying to work out if neli (in its current state) can be used for tracing proc_events using the netlink Connector protocol. I can't find any mention of the types similar to those declared in 'linux/cn_proc.h' used by netlink Connector programs written in C.

Apologies in advance if my question is completely out of place. I'm hoping neli can help me avoid bindgen and wrangling unsafe c wrappers.

Bben01:
Close #275 

So, a very early stage of the PR, but a working connector.

A few things to notice:

1. I commented the payload parsing because currently I use the `Nlmsghdr` to get the response from the socket, but with the connector protocol, a valid response comes with `nl_type` of Done, so the Payload parsed it as an error
2. In CnMsg, I flatten the struct `cb_id` from the kernel, but I don't really know if I should do it or not 

```c
struct cb_id {
	__u32 idx;
	__u32 val;
};

struct cn_msg {
	struct cb_id id;

	__u32 seq;
	__u32 ack;

	__u16 len;		/* Length of the following data */
	__u16 flags;
	__u8 data[];
};
```

To:

```rust
#[derive(
    Builder, Getters, Clone, Debug, PartialEq, Eq, Size, ToBytes, FromBytesWithInput, Header,
)]
#[neli(from_bytes_bound = "P: Size + FromBytesWithInput<Input = usize>")]
#[builder(build_fn(skip))]
#[builder(pattern = "owned")]
pub struct CnMsg<P> {
    /// Index of the connector (idx)
    #[getset(get = "pub")]
    idx: u32,
    /// Value (val)
    #[getset(get = "pub")]
    val: u32,
    /// Sequence number
    #[getset(get = "pub")]
    seq: u32,
    /// Acknowledgement number
    #[getset(get = "pub")]
    ack: u32,
    /// Length of the payload
    #[builder(setter(skip))]
    #[getset(get = "pub")]
    len: u16,
    /// Flags
    #[getset(get = "pub")]
    flags: u16,
    /// Payload of netlink message
    #[neli(size = "len as usize")]
    #[neli(input = "(len as usize)")]
    #[getset(get = "pub")]
    pub(crate) payload: P,
}
```
3. Currently I only support ProcEvents from the connector, but it can actually have different payloads, based on the idx and val. 
4. I need to check that idx and val are correct, otherwise I need to return an error
5. I don't really know if the parsing of the response is good enough or should be refactored, but from what I can understand, it cannot be automatically derived (the C code contains a number + union (instead of a more idiomatic enum in rust)
6. I still need to define al the consts, but I will do it after the PR in libc lands (https://github.com/rust-lang/libc/pull/4434)
7. Better docs + example

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ bitflags = "2.4"
 byteorder = "1.3"
 derive_builder = "0.20"
 getset = "0.1.2"
-libc = "0.2.172"
+libc = "0.2.174"
 log = "0.4"
 
 [dependencies.neli-proc-macros]
--- a/examples/procmon.rs
+++ b/examples/procmon.rs
@@ -1,0 +1,87 @@
+//! A small process monitor example using netlink connector messages.
+//!
+//! You must run this example with root privileges, in the root pid namespace
+//!
+//! See this blog post for more details:
+//! https://nick-black.com/dankwiki/index.php/The_Proc_Connector_and_Socket_Filters
+
+use neli::{
+    connector::{CnMsg, ProcEvent, ProcEventHeader},
+    consts::{
+        connector::{CnMsgIdx, CnMsgVal, ProcCnMcastOp},
+        nl::{NlmF, Nlmsg},
+        socket::NlFamily,
+    },
+    nl::{NlPayload, NlmsghdrBuilder},
+    socket::synchronous::NlSocketHandle,
+    utils::Groups,
+};
+use std::fs;
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    env_logger::init();
+
+    let pid = std::process::id();
+    let socket = NlSocketHandle::connect(
+        NlFamily::Connector,
+        Some(pid),
+        Groups::new_bitmask(CnMsgIdx::Proc.into()),
+    )?;
+
+    let subscribe = NlmsghdrBuilder::default()
+        .nl_type(Nlmsg::Done)
+        .nl_flags(NlmF::empty())
+        .nl_pid(pid)
+        .nl_payload(NlPayload::Payload(
+            neli::connector::CnMsgBuilder::default()
+                .idx(CnMsgIdx::Proc)
+                .val(CnMsgVal::Proc)
+                .payload(ProcCnMcastOp::Listen)
+                .build()?,
+        ))
+        .build()?;
+
+    socket.send(&subscribe)?;
+
+    loop {
+        for event in socket.recv::<Nlmsg, CnMsg<ProcEventHeader>>()?.0 {
+            let ProcEvent::Exec { process_pid, .. } = event?
+                .get_payload()
+                .ok_or("Failed to extract payload")?
+                .payload()
+                .event
+            else {
+                continue;
+            };
+
+            let exe = fs::read_link(format!("/proc/{}/exe", process_pid))
+                .map(|p| p.display().to_string())
+                .unwrap_or_else(|_| "unknown".to_string());
+            let cmdline = cmdline_to_string(process_pid).unwrap_or_else(|_| "unknown".to_string());
+            println!(
+                "Process created: PID: {}, Exe: {}, Cmdline: {}",
+                process_pid, exe, cmdline
+            );
+        }
+    }
+}
+
+fn cmdline_to_string(pid: i32) -> std::io::Result<String> {
+    // 1) Read the entire file into a byte‐buffer in one go
+    let mut data = fs::read(format!("/proc/{pid}/cmdline"))?;
+
+    // 2) In‐place map all remaining NULs to spaces
+    for b in &mut data {
+        if *b == 0 {
+            *b = b' ';
+        }
+    }
+
+    // 3) SAFELY convert bytes → String without re‐checking UTF‐8
+    //
+    //    We can do this because /proc/cmdline is guaranteed to be valid UTF-8
+    //    on Linux (arguments must be passed as UTF-8), so we skip the cost
+    //    of a UTF-8 validation pass.
+    let s = unsafe { String::from_utf8_unchecked(data) };
+    Ok(s)
+}
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -1,0 +1,340 @@
+//! Connector module for Linux Netlink connector messages.
+//!
+//! This module provides support for the Linux Netlink connector subsystem,
+//! which creates a communication channel between userspace programs and the kernel.
+//! It allows applications to receive notifications about various kernel events.
+//!
+//! This module currently provides full support for the Linux proc connector protocol,
+//! enabling the reception and handling of process lifecycle events such as creation,
+//! termination, exec, UID/GID/sid changes, tracing, name changes, and core dumps.
+//!
+//! ## Supported protocols
+//! At this time, only the proc connector (`PROC_CN`) protocol is fully implemented.
+//!
+//! ## Extensibility
+//! The implementation can be extended in two ways:
+//! 1. By defining additional types and logic in your own crate and using them with this module.
+//! 2. By using a `Vec<u8>` as a payload and manually parsing protocol messages to suit other connector protocols.
+//!
+//! This design allows both high-level ergonomic handling of proc events and low-level manual parsing for custom needs.
+use crate::{
+    self as neli,
+    consts::connector::{CnMsgIdx, CnMsgVal, ProcEventType},
+    err::{DeError, MsgError},
+    FromBytesWithInput, Header, Size, ToBytes,
+};
+use byteorder::{NativeEndian, ReadBytesExt};
+use derive_builder::Builder;
+use getset::Getters;
+use log::trace;
+use std::{io::Cursor, io::Read};
+
+/// Netlink connector message header and payload.
+#[derive(
+    Builder, Getters, Clone, Debug, PartialEq, Eq, Size, ToBytes, FromBytesWithInput, Header,
+)]
+#[neli(from_bytes_bound = "P: Size + FromBytesWithInput<Input = usize>")]
+#[builder(pattern = "owned")]
+pub struct CnMsg<P: Size> {
+    /// Index of the connector (idx)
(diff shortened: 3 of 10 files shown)


jbaublitz:
First of all, thank you so much for your work on this! I really appreciate the time you've put into this.

> Close #275
> 
> So, a very early stage of the PR, but a working connector.
> 
> A few things to notice:
> 
>     1. I commented the payload parsing because currently I use the `Nlmsghdr` to get the response from the socket, but with the connector protocol, a valid response comes with `nl_type` of Done, so the Payload parsed it as an error\

I've recently touched this code and I'll have to review it, but I would expect a payload of `Empty` if there is an error code of 0.

> 
>     2. In CnMsg, I flatten the struct `cb_id` from the kernel, but I don't really know if I should do it or not
> 
> 
> ```c
> struct cb_id {
> 	__u32 idx;
> 	__u32 val;
> };
> 
> struct cn_msg {
> 	struct cb_id id;
> 
> 	__u32 seq;
> 	__u32 ack;
> 
> 	__u16 len;		/* Length of the following data */
> 	__u16 flags;
> 	__u8 data[];
> };
> ```
> 
> To:
> 
> ```rust
> #[derive(
>     Builder, Getters, Clone, Debug, PartialEq, Eq, Size, ToBytes, FromBytesWithInput, Header,
> )]
> #[neli(from_bytes_bound = "P: Size + FromBytesWithInput<Input = usize>")]
> #[builder(build_fn(skip))]
> #[builder(pattern = "owned")]
> pub struct CnMsg<P> {
>     /// Index of the connector (idx)
>     #[getset(get = "pub")]
>     idx: u32,
>     /// Value (val)
>     #[getset(get = "pub")]
>     val: u32,
>     /// Sequence number
>     #[getset(get = "pub")]
>     seq: u32,
>     /// Acknowledgement number
>     #[getset(get = "pub")]
>     ack: u32,
>     /// Length of the payload
>     #[builder(setter(skip))]
>     #[getset(get = "pub")]
>     len: u16,
>     /// Flags
>     #[getset(get = "pub")]
>     flags: u16,
>     /// Payload of netlink message
>     #[neli(size = "len as usize")]
>     #[neli(input = "(len as usize)")]
>     #[getset(get = "pub")]
>     pub(crate) payload: P,
> }
> ```
> 

I calculated the alignment and my only concern here would be on architectures other than x86. For x86, this looks reasonable.

>     3. Currently I only support ProcEvents from the connector, but it can actually have different payloads, based on the idx and val.
> 
>     4. I need to check that idx and val are correct, otherwise I need to return an error
> 
>     5. I don't really know if the parsing of the response is good enough or should be refactored, but from what I can understand, it cannot be automatically derived (the C code contains a number + union (instead of a more idiomatic enum in rust)

Can you give me a little bit more detail here? There may still be a way to make this work.

> 
>     6. I still need to define al the consts, but I will do it after the PR in libc lands ([Add constants from linux/cn_proc.h and linux/connector.h rust-lang/libc#4434](https://github.com/rust-lang/libc/pull/4434))
> 
>     7. Better docs + example

Bben01:
> I've recently touched this code and I'll have to review it, but I would expect a payload of Empty if there is an error code of 0.

I don't understand what you mean. What I am saying is that the `Nlmsghdr.nl_type` of the response in the connector protocol will have a value of 3 (`Nlmsg::Done`), so when the payload is parsed, on this line (`let ty_const: u16 = input_type.into();`), the value will be `Nlmsg::Done`, and that will then go to the first branch in the next if, so it won't parse the payload correctly.

> I calculated the alignment and my only concern here would be on architectures other than x86. For x86, this looks reasonable.

So, should I change it or not?

> Can you give me a little bit more detail here? There may still be a way to make this work.

The connector protocol is used as a generic kernel to user space communication, each type has its own payload.

Currently, these are the different types that the kernel supports:

```rust
pub const CN_IDX_PROC: c_uint = 0x1;
pub const CN_VAL_PROC: c_uint = 0x1;
pub const CN_IDX_CIFS: c_uint = 0x2;
pub const CN_VAL_CIFS: c_uint = 0x1;
pub const CN_W1_IDX: c_uint = 0x3;
pub const CN_W1_VAL: c_uint = 0x1;
pub const CN_IDX_V86D: c_uint = 0x4;
pub const CN_VAL_V86D_UVESAFB: c_uint = 0x1;
pub const CN_IDX_BB: c_uint = 0x5;
pub const CN_DST_IDX: c_uint = 0x6;
pub const CN_DST_VAL: c_uint = 0x1;
pub const CN_IDX_DM: c_uint = 0x7;
pub const CN_VAL_DM_USERSPACE_LOG: c_uint = 0x1;
pub const CN_IDX_DRBD: c_uint = 0x8;
pub const CN_VAL_DRBD: c_uint = 0x1;
pub const CN_KVP_IDX: c_uint = 0x9;
pub const CN_KVP_VAL: c_uint = 0x1;
pub const CN_VSS_IDX: c_uint = 0xA;
pub const CN_VSS_VAL: c_uint = 0x1;
```

I've only worked with proc connectors, so I do not know how the other works, but I suspect it won't be too different. 

I could maybe have another option that won't parse the payload if the type is unknown, and instead will keep the raw bytes.

jbaublitz:
> > I've recently touched this code and I'll have to review it, but I would expect a payload of Empty if there is an error code of 0.
> 
> I don't understand what you mean. What I am saying is that the `Nlmsghdr.nl_type` of the response in the connector protocol will have a value of 3 (`Nlmsg::Done`), so when the payload is parsed, on this line (`let ty_const: u16 = input_type.into();`), the value will be `Nlmsg::Done`, and that will then go to the first branch in the next if, so it won't parse the payload correctly.
> 

Ah, now I understand what you mean. Can you outline what payload you expect in the `Done` packet? I'm happy to change the parsing a bit to be a bit more flexible.

> > I calculated the alignment and my only concern here would be on architectures other than x86. For x86, this looks reasonable.
> 
> So, should I change it or not?

I'm currently asking some colleagues more knowledgeable on architecture than I am about this. In the Fedora space, @decathorpe and @sbrivio-rh opened an issue about s390x so they may have thoughts here.

> 
> > Can you give me a little bit more detail here? There may still be a way to make this work.
> 
> The connector protocol is used as a generic kernel to user space communication, each type has its own payload.
> 
> Currently, these are the different types that the kernel supports:
> 
> ```rust
> pub const CN_IDX_PROC: c_uint = 0x1;
> pub const CN_VAL_PROC: c_uint = 0x1;
> pub const CN_IDX_CIFS: c_uint = 0x2;
> pub const CN_VAL_CIFS: c_uint = 0x1;
> pub const CN_W1_IDX: c_uint = 0x3;
> pub const CN_W1_VAL: c_uint = 0x1;
> pub const CN_IDX_V86D: c_uint = 0x4;
> pub const CN_VAL_V86D_UVESAFB: c_uint = 0x1;
> pub const CN_IDX_BB: c_uint = 0x5;
> pub const CN_DST_IDX: c_uint = 0x6;
> pub const CN_DST_VAL: c_uint = 0x1;
> pub const CN_IDX_DM: c_uint = 0x7;
> pub const CN_VAL_DM_USERSPACE_LOG: c_uint = 0x1;
> pub const CN_IDX_DRBD: c_uint = 0x8;
> pub const CN_VAL_DRBD: c_uint = 0x1;
> pub const CN_KVP_IDX: c_uint = 0x9;
> pub const CN_KVP_VAL: c_uint = 0x1;
> pub const CN_VSS_IDX: c_uint = 0xA;
> pub const CN_VSS_VAL: c_uint = 0x1;
> ```
> 
> I've only worked with proc connectors, so I do not know how the other works, but I suspect it won't be too different.
> 
> I could maybe have another option that won't parse the payload if the type is unknown, and instead will keep the raw bytes.

A pattern that may be helpful to look at for this is the attributes. I also kept the bytes and allowed parsing after the fact. That seems to have worked pretty well so far. Unfortunately, it's not quite as intuitive as just encoding it in a type parameter, but it's flexible and generally works well.

jbaublitz:
> > Can you outline what payload you expect in the Done packet?
> 
> You can look at the example I added, in the recv call

Will do.

> 
> > A pattern that may be helpful to look at for this is the attributes
> 
> Could you clarify this point? I think I didn't understand fully what you meant

The generic netlink attributes (`Nlattr`) and routing attributes (`Rtattr`) follow a similar pattern to what you're describing here. You may be able to find some inspiration in the handling code for those data structures if you're stuck.

Bben01:
You're welcome

I think you should look at the (small) change of the payload parsing in `nl.rs` before releasing v0.7.0 because the fix could potentially be breaking

jbaublitz:
Do you mean the change in this PR?

Bben01:
Yes

jbaublitz:
> Fedora

@decathorpe Thanks for the offer. Currently QEMU s390x seems to be working better than I was expecting. Very slow but doable for what I need it for.

jbaublitz:
@Bben01 Overall, I am happy with the code, but will provide a more detailed code review after I release 0.7.0. I tested your example code and the example code for DUMP directives and, while I will probably implement it differently in the 0.7.0 version, I see why you made the choice you did and I will try to preserve it in a way that doesn't require reparsing. I believe this format is slightly different due to the subscription model. It seems like that is why each packet is of type DONE and contains a payload of the expected type. Unfortunately, what I don't know is how this maps to other subscription-like netlink subsystems. I'm going to do some reading and try to see if I can find a similar protocol but honestly I would have expected an API like this to be exposed as a multicast group. That is the typical (and generally blessed way) of subscribing to a notification stream. I do support that in this library but unfortunately it seems like that is not the way it's implemented for this protocol in the kernel.

Bben01:
Thanks, I cleaned up the git history, and fixed the new clippy lints (two separate commits)

jbaublitz:
Thanks so much for this contribution! I believe the clippy errors are unrelated to your code, so I'm going to merge.

PR #278, opened 2025-06-11
Fix rta deserialization test for big endian

### LINKED ISSUE #273: test::test_rta_deserialize fails on s390x (on big-endian in general, I suppose)
**Version of neli**
0.7.0-rc3

**Describe the bug**
test::test_rta_deserialize fails on s390x. See: https://kojipkgs.fedoraproject.org//work/tasks/7795/130447795/build.log (first build on Fedora, I'm the maintainer for the new package).

**To Reproduce**
Run the tests on big-endian, I suppose.

---

I'm currently looking into this, no obvious clues yet. There's probably a byteswap too many somewhere.

jbaublitz:
@sbrivio-rh I won't merge until I've confirmed I've solved all of your issues. Can you confirm that this at least resolves the test failure that it targets? I'm happy to do testing on my end to if you can point me to how I can run it.

### DIFF
--- a/src/rtnl.rs
+++ b/src/rtnl.rs
@@ -428,6 +428,8 @@ mod test {
 
     use std::net::Ipv4Addr;
 
+    use byteorder::{NativeEndian, WriteBytesExt};
+
     use crate::{
         consts::{nl::NlmF, socket::NlFamily},
         err::RouterError,
@@ -441,8 +443,11 @@ mod test {
     fn test_rta_deserialize() {
         setup();
 
-        let buf = &4u32.to_ne_bytes() as &[u8];
-        Rtattr::<Rta, Buffer>::from_bytes(&mut Cursor::new(buf)).unwrap();
+        let mut buf = Cursor::new(vec![]);
+        buf.write_u16::<NativeEndian>(4).unwrap();
+        buf.write_u16::<NativeEndian>(0).unwrap();
+        buf.set_position(0);
+        Rtattr::<Rta, Buffer>::from_bytes(&mut buf).unwrap();
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


sbrivio-rh:
Sorry, I didn't have a chance to try this out yet, but I should be able to within the next couple of days.

jbaublitz:
@sbrivio-rh No worries! I really want to get the release out as soon as possible so I think I'm going to merge this for now as I'm confident it's correct. I'm going to leave #273 open for the time being and return to it when I'm not trying to get a release out. If you want to more interactively debug it on IRC as some point, I'm happy to help!

sbrivio-rh:
Finally running scratch build now with this fix cherry-picked on top of 0.7.0-rc3 and tests on s390x, https://koji.fedoraproject.org/koji/taskinfo?taskID=134182890. I'll check in a while.

sbrivio-rh:
...so, yes, I can confirm that this fixes `test_rta_deserialize`.

Tests as a whole still fail on s390x for some reason I couldn't quite grasp yet. I ran the same scratch build twice, here are the s390x build logs:

- https://kojipkgs.fedoraproject.org//work/tasks/2955/134182955/build.log ...tests seem to fail but none of the single tests do
- https://kojipkgs.fedoraproject.org//work/tasks/6545/134186545/build.log here we finally get the good old `memory allocation of 34359738372 bytes failed`

Note that I tested 0.7.0-rc3 with just this commit on top (see source RPM if you want to double check things).

To run these builds, while it's probably not the most efficient way, here is what I did:
- prepare the patch from the `neli` repository:
  - I checked out 0.7.0-rc3 (`git checkout neli-v0.7.0-rc3`)
  - cherry-picked the commit (`git cherry-pick 24d618557bb264fedeebdc75db83ca13db90a54e`)
  - then did `git format-patch -1` and copied it to `~/rpmbuild/SOURCES/`
- `git clone https://src.fedoraproject.org/rpms/rust-neli.git`
- go there and edit the spec file to include the patch we just obtained from `git format-patch`, say:
```
Patch:          0001-Fix-rta-deserialization-test-for-big-endian.patch
```
- also re-enable the build for s390x by adding it back to `supported_arches`
- `fedpkg scratch-build --srpm`. For this, you'll need to be authenticated on fedoraproject.org with `kinit` or `fkinit`, and have `fedpkg` and dependencies installed

...let me know if there's something else I should test. I can also add patches all the way to the current HEAD if you think it makes sense.

PR #279, opened 2025-06-13
Directly implement FromBytesWithInput and ToBytes for Buffer

chrspeich:
Currently, Buffer will forward the implementations of FromBytesWithInput and ToBytes to the generic implementation of Vec<_>. Is implementation is quite inefficient when the contained element is small (like u8) and there might be many elements (like a Buffer).

I don't see any way to specialise the Vec<u8> case, so I've implemented FromBytesWithInput and ToBytes directly for Buffer.

A very simple benchmark with 1500 bytes show these results:
 - From Buffer: 27.94ns/call
 - From Vec<u8>: 862.63ns/call
 - To Buffer: 28.34ns/call
 - To Vec<u8>: 1288.43ns/call

This shows a massive improvement when Buffer is used.

Additionally, I've added Into<Vec<u8>> to Buffer to easily retrieve the stored Vec.

---

Most of the time netlink is only used for slow actions. However sometimes, there can be quite large and many packets. One example ist mac80211_hwsim which uses netlink to transfer full wlan frames from/to userspace and the performance of Buffer then matters.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -8,6 +8,7 @@
 
 use std::{
     fmt::{self, Debug},
+    io::{Read, Write},
     iter::FromIterator,
     slice::{Iter, IterMut},
 };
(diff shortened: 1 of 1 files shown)


jbaublitz:
@chrspeich Thanks for this improvement! This has been on my radar and I think this sounds like a reasonable solution. I'll review it today and if it looks good, I'll merge it!

src/types.rs
```diff
@@ -24,9 +25,35 @@ use crate::{
 };
 
 /// A buffer of bytes.
-#[derive(Clone, PartialEq, Eq, Size, FromBytesWithInput, ToBytes)]
+#[derive(Clone, PartialEq, Eq, Size)]
 pub struct Buffer(#[neli(input)] Vec<u8>);
```
jbaublitz, line 29: If you're overriding the derive implementations, please remove the inner attribute meant for those macros.

jbaublitz:
@chrspeich It appears one of the PR revisions is addressed by the clippy CI. There's one additional clippy error to take a look at. I think it's specifically asking for use of `write_all`.

PR #280, opened 2025-06-13
Pass full io::Error to user

chrspeich:
Currently, for io::Error only the error kind is returned to the user. This is a lossy conversion and may leave the user with a Uncategorized error and no way to debug this further.

As io::Error is not cloneable, wrap it in an Arc.

---

I'm not sure if there was an deliberate decision behin only keeping the ErrorKind other than io::Error not being Cloneable. This is a user facing API change, but I hope this is okay in an rc.

In my case I've got a Uncategorized error in `recv` which was actually a `ENOBUFS`. This error means that I'm not reading from the netlink socket fast enough. Without this patch there is no way I would have diagnosed the problem, so I hope the API-Change is acceptable.

### DIFF
--- a/src/err.rs
+++ b/src/err.rs
@@ -21,6 +21,7 @@ use std::{
     io::{self, Cursor, ErrorKind},
     str::Utf8Error,
     string::FromUtf8Error,
+    sync::Arc,
 };
 
 use derive_builder::{Builder, UninitializedFieldError};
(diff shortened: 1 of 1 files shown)


jbaublitz (approved):
Overall this looks like a superior solution, thanks for the contribution! Tests all seem to be passing and I don't have any concerns about the code changes so I'll include this in the 0.7.0 release.

PR #288, opened 2025-06-29
Add option to set recv buffer size

### LINKED ISSUE #230: Setting socket buffer sizes
I've been looking through some C examples for some SocketCAN Netlink communications, and they all seem to set the socket buffer sizes. Like:

```
int sndbuf = 32768;
int rcvbuf = 1048576;

setsockopt(fd, SOL_SOCKET, SO_SNDBUF, (void *)&sndbuf, sizeof(sndbuf));
setsockopt(fd, SOL_SOCKET, SO_RCVBUF, (void *)&rcvbuf, sizeof(rcvbuf));
```

Is there a way to do this easily with this library - other than getting the socket file handle and doing it manually with libc functions? If it doesn't already exist, I can give it a try and send over a PR.

Bben01:
Closes #230 

I think that this was the intent of the issue author.
I've read a bit about the option, and it seems that we cannot rely on the kernel to set a specific size, so we cannot rely on that to simplify the logic in neli's buffers.

I didn't provide an option to set the send buffer size, because the default should be good enough (netlink isn't typically used to send large amounts of data).

### DIFF
--- a/src/socket/asynchronous.rs
+++ b/src/socket/asynchronous.rs
@@ -159,6 +159,26 @@ impl NlSocketHandle {
         Ok((vec, groups))
     }
 
+    /// Set the size of the receive buffer for the socket.
+    ///
+    /// This can be useful when communicating with a service that sends a high volume of
+    /// messages (especially multicast), and your application cannot process them fast enough,
+    /// leading to the kernel dropping messages. A larger buffer may help mitigate this.
+    ///
+    /// The value passed is a hint to the kernel to set the size of the receive buffer.
+    /// The kernel will double the value provided to account for bookkeeping overhead.
+    /// The doubled value is capped by the value in `/proc/sys/net/core/rmem_max`.
+    ///
+    /// The default value is `/proc/sys/net/core/rmem_default`
+    ///
+    /// See `socket(7)` documentation for `SO_RCVBUF` for more information.
+    pub fn set_recv_buffer_size(&self, size: usize) -> Result<(), SocketError> {
+        self.socket
+            .get_ref()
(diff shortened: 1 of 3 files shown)


jbaublitz:
@Bben01 I had one additional question. What are your thoughts on using `rmem_max` as the maximum value for the `BufferPool` implementation?

Bben01:
That could work, although there is also another flag `SO_RCVBUFFORCE` that will override the max value (only for root).

I could try to send a PR for that

Bben01:
After thinking a bit about the change, I don't think its wise anymore:
1. The limit appears to be at different paths for each OS, from what I can tell, only linux supports net.core.rmem_max, other OS will have another variable name for that
2. We could take the socket and use getsockopt(SO_RCVBUF), but then each OS report the size in a different manner: linux double the value, others may or may not double it
3. We have to turn the now constified at compile time function to a syscall

About the buffer pool, is there a reason that the default parallel is 3?

Bben01:
Hi @jbaublitz,

I’m not sure if you received a notification for my previous comment.

Also, when do you plan to release 0.7.1?

jbaublitz:
@Bben01 I did receive your comment but was away from my computer for a bit on vacation. I'm fine just leaving it the way it is rather than trying to engineer something that may break on different platforms. As for the release, are you looking to get 0.7.1 out as quickly as possible? I can bump some of the desired features back to 0.7.2 if it's urgent and release today.

Bben01:
@jbaublitz ok, no problem

It's not really urgent, but it would be nice (if it's not too much work)

PR #289, opened 2025-06-29
Add const generic implementations for FromBytes and ToBytes traits for arrays

### LINKED ISSUE #210: Support FromBytes for [u8; N]
No

Some attributes return fixed length byte arrays.  (eg a mac address, [u8;6], or the 802.11 VHT MCS Set [u8;32] etc).  Some types (eg IP Addresses) have conversions from fixed sized arrays, but not from slices.

I would like `attr.get_payload_as::<[u8;32]>()` to work, which I think is approximately:

```rust
impl<const N: usize> FromBytes for [u8;N] { ... }
impl<const N: usize> ToBytes for [u8;N] { ... }
impl<const N: usize> Size for [u8;N] { ... }
```

(I don't understand the difference between `FromBytes` and `FromBytesWithInput`, maybe this should be the WithInput variant?  I dunno.  This also means I don't understand when to use `.get_payload_as()` vs `.get_payload_as_with_len()`)

It might also be possible (although I imagine a bit tricker to implement in an efficient manner) to have:

```rust
impl<const N: usize, T: Size + FromBytes> FromBytes for [T; N] { ... }
impl<const N: usize, T: Size + ToBytes> ToBytes for [T;N] { ... }
impl<const N: usize, T: Size + FromBytes> Size for [T; N] { ... }
```

which would allow for arbitrary arrays of u8, u16, u32 and even more complex types.  (Interface counters are often returned as arrays of u32 or u64 for instance).

Currently I get the reference to the slice, then use TryFrom to convert it to the correct sized array but this includes an extra error type that couldn't occur if get_payload_as() had returned the correct size.

Bben01:
Closes #210 

Since I've already implemented `Size` for `[u8; N]`, I think that we could also implement it for the other traits to make the crate smoother to use

About implementing them for `[T; N]`, I don't really know what would be the use case, so maybe not implementing it until someone ask for it is the right choice

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -442,6 +442,14 @@ impl FromBytesWithInput for String {
     }
 }
 
+impl<const N: usize> FromBytes for [u8; N] {
+    fn from_bytes(buffer: &mut Cursor<impl AsRef<[u8]>>) -> Result<Self, DeError> {
+        let mut arr = [0u8; N];
+        buffer.read_exact(&mut arr)?;
+        Ok(arr)
+    }
+}
+
 impl Size for &'_ [u8] {
     fn unpadded_size(&self) -> usize {
         self.len()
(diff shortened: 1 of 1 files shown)


jbaublitz:
Thanks for all of your help. I'll try to take a look at this tomorrow.

jbaublitz (approved):

PR #290, opened 2025-07-01
Expand nl80211 example

a-gavin:
Expand the nl80211 example to include interface-level information printouts.

~~This is somewhat self-motivated, as I am having trouble with parsing the SSID field. Unlike interface and wiphy name attributes, the SSID is not NULL-terminated (note the partial SSID in the printout, should be `partialssid`). I'm sure with fresh eyes I can find it, but doesn't hurt to update the example haha!~~ See below comment on using `get_payload_as_with_len_borrowed()` with `&[u8]`.

Note that P2P device interfaces do not have an interface index or name when queried with `NL80211_CMD_GET_INTERFACE`, at least how I've updated the example.

```Bash
$ cargo run --example nl80211 --features async
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/examples/nl80211`
WiFi radios:
Wiphy:      0
WiphyName:  phy0


WiFi interfaces:
Wiphy:      0
Iftype:     P2pDevice
Wdev:       3

Ifname:     wlp3s0
Wiphy:      0
Iftype:     Station
Wdev:       1
Ssid:       partialssi
```

### DIFF
--- a/examples/nl80211.rs
+++ b/examples/nl80211.rs
@@ -1,4 +1,4 @@
-use std::error::Error;
+use std::{error::Error, iter::once};
 
 #[cfg(feature = "async")]
 use neli::router::asynchronous::NlRouter;
@@ -10,8 +10,10 @@ use neli::{
         nl::{GenlId, NlmF},
         socket::NlFamily,
     },
+    genl::{AttrTypeBuilder, NlattrBuilder},
     genl::{Genlmsghdr, GenlmsghdrBuilder, NoUserHeader},
     nl::{NlPayload, Nlmsghdr},
+    types::GenlBuffer,
     utils::Groups,
 };
 
@@ -19,6 +21,7 @@ use neli::{
 pub enum Nl80211Command {
     Unspecified = 0,
     GetWiPhy = 1,
+    GetInterface = 5,
     /* Many many more elided */
 }
 impl neli::consts::genl::Cmd for Nl80211Command {}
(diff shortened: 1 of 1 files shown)


a-gavin:
Updated PR with fixed SSID printout. Aside from adding additional info to the `nl80211.rs` example, these changes now include usage of the `get_payload_as_with_len_borrowed()` funciton with a `&[u8]` byte array. No other examples use the function, so it should hopefully serve as a good reference to others.

Also, as a heads up, I'm slowly working on some `nl80211` constants in my personal tree [here](https://github.com/a-gavin/neli/tree/dev) (and some `mac80211_hwsim` constants [here](https://github.com/a-gavin/neli/tree/mac80211_hwsim)). I have a PR open to Rust `libc` crate [here](https://github.com/rust-lang/libc/pull/4546) to add a first batch of `nl80211` constants (`mac80211_hwsim` and more `nl80211` constants to come). Will PR to this tree when things land.

jbaublitz:
@a-gavin I'm happy to address the issues you've raised with CI and find a better way to point to the kernel headers that works cross platform. However, the reason I pointed you to that branch is that I do not want to release a separate crate for bindings. I've worked on a lot of bindings like https://github.com/stratis-storage/libcryptsetup-rs and https://github.com/stratis-storage/libblkid-rs and from my vantage point, often the benefit of releasing a separate crate is when you would like to logically separate the high-level and low-level bindings. In the case of a lot of autogenerated function bindings, this can be useful, but the tradeoff is that often development and release work becomes slightly more complicated. I do not feel the benefits outweigh the negatives here, especially when we're only generating bindings for constants as all of the function-based work is already covered by the libc crate. You can feel free to try to address your concerns in your solution, but I do not want a separate crate for the bindings in this case and I've drafted a rough example of how that could be achieved in #292.

a-gavin:
@jbaublitz Sounds good. I believe this PR is ready for merge and close if you're okay with merging only example changes changes. Will do other items in separate PRs going forward.

jbaublitz (approved):
Thank you for the thorough documentation! I'll merge this.

PR #292, opened 2025-07-22
Add bindgen capability in neli

jbaublitz:
Related to #290


jbaublitz:
It looks like this will not be necessary after further discussion in #290

PR #295, opened 2025-07-23
neli-0.7.1 and neli-proc-macros-0.2.1

jbaublitz:


jbaublitz:
@Bben01 v0.7.1 is released.

PR #298, opened 2025-09-02
Added Rtm::Getstats and Rtm::Newstats

eternalfrustation:
Basically the title, since these were missing, 2 constants

### DIFF
--- a/src/consts/rtnl.rs
+++ b/src/consts/rtnl.rs
@@ -249,6 +249,8 @@ pub enum Rtm {
     Newnsid = libc::RTM_NEWNSID,
     Delnsid = libc::RTM_DELNSID,
     Getnsid = libc::RTM_GETNSID,
+    Newstats = libc::RTM_NEWSTATS,
+    Getstats = libc::RTM_GETSTATS,
 }
 
 /// Enum usable with [`Rtattr`][crate::rtnl::Rtattr] field,


eternalfrustation:
Since initial commit, i have also added filter-mask attributes for IFLA_STATS, and the corresponding msg struct.

eternalfrustation:
I am sorry, i did not receive a notification for this PR.
I don't really know what kind of test this would need.

eternalfrustation:
> Just one last thing to check. If that looks good I'll merge this.

These constants are available in linux-sys-raw.
If needed, i can introduce linux-sys-raw as a dependency, but it feels like only a core maintainer should hold the right for that.

Also, i can't make any changes anyways for about 2 days.

jbaublitz:
No rush, sorry I haven't been able to look at this for a bit.

I'm happy to add that as a dependency as it appears a large number of constants that users have been requesting recently are not available in libc and libc has not wanted to accept the constants upstream I think due to their scope not being strictly related to libc.

jbaublitz:
@eternalfrustation If you could rebase onto main and resolve conflicts, it should already be added by another PR.

jbaublitz:
@eternalfrustation No, please rebase and undo the merge. I want to keep the git history clean. Once you rebase and modify the constants to use linux-sys-raw, I'll do a full review and likely merge unless there are any major problems. I also intend to put out a release after I merge this PR just to make the last two PRs merged available.

Issue #301, opened 2025-10-07
Support `NlattrType`s for generic unsigned integer types (Only `u16` presently)


a-gavin:
**Would adding this feature cause a breaking change?**
I'm not sure, but hopefully no?

**Is your feature request related to a problem? Please describe.**
`NlattrType` impl is only supported for `u16` serialized type enums. However, there are several non-`u16` attributes in the nl80211 generic Netlink specification (all `u32` that I can tell from work I've done recently).

While defining these actually `u32` attributes as `u16` may permit usage for parsing messages received from the kernel module (e.g. nested attributes), this workaround can prevent usage when sending messages to the kernel module.

I have specifically seen this when defining/using the nested `enum nl80211_iftype` `u32` attr enum. When attempting to add a new interface with the `u16` workaround, the kernel returns an error message complaining about invalid length. Wireshark netlink capture confirms this and working case using the `iw` tool shows success with the correct length.

**Describe the solution you'd like**
Added support for arbitrary unsigned integer type serialized type enums, specifically `u32`.

**Additional context**
Example where the actual type is `u32`, but only defining as `u16` will compile:
```Rust
/// Virtual interface types
// TODO: This is actually a u32, but compiler won't allow anything but u16 for Nlattr
#[neli_enum(serialized_type = "u16")]
pub enum Nl80211IfType {
    /// Unspecified type, driver decides
    Unspecified = nl80211_iftype::NL80211_IFTYPE_UNSPECIFIED as u16,
    ...
    /// NAN device interface type (not a netdev)
    Nan = nl80211_iftype::NL80211_IFTYPE_NAN as u16,
}
impl neli::consts::genl::NlAttrType for Nl80211IfType {}
```

a-gavin:
Curious what your thoughts on this are. Happy to help with a solution where I can.

I ran into this again with another nl80211 attribute i'd like to parse out, which is a variation on the theme. The `NL80211_ATTR_SUPPORTED_COMMANDS` returned during a `iw phy0 info` command contains a nested array of `NL80211_CMD_X`s.

jbaublitz:
@a-gavin The problem has less to do with proc macros and more the design of the library. For example, when we implement ToBytes for a large struct, the idea is that every single field must also implement ToBytes. ToBytes specifically has exactly one serialized representation and this allows users who know nothing about the serialization to simply call .to_bytes() and get the binary representation that they expect.

[...]

PR #302, opened 2025-10-07
Implement nl80211 and mac80211_hwsim generic Netlink constants

a-gavin:
Something to chew on while I continue to add more to the `neli`-side of the `nl80211` constants for a project I'm working on. I expect to add several more enums over the next month or so to the `nl80211` definitions (`mac80211_hwsim` should be okay as is, pending any feedback).

Given this and [this blocking feature request/issue](https://github.com/jbaublitz/neli/issues/301) I raised regarding `NlattrType` serialized types, I'm raising this as a draft. For a finalized PR, I plan to properly document the the `nl80211` constants, update the existing `nl80211` example to use the constants, and of course update the commit message.

For some quick context, the `nl80211` generic Netlink interface is the primary method used to configure and query WiFi interfaces in the Linux kernel (e.g. `wpa_supplicant`/`hostapd`, `iwd`, `iw`). The `mac80211_hwsim` generic Netlink Interface is the communication method used to configure and query virtual WiFi radios in the Linux kernel and is used by several projects for virtualized testing (e.g. `wpa_supplicant`/`hostapd`, `iwd`).

To grab `nl80211` definitions, this PR uses the `linux-raw-sys` crate as [previously discussed](https://github.com/jbaublitz/neli/pull/290#issuecomment-3113672917). Support for all `nl80211` Linux 6.16 constants landed in the `linux-raw-sys` crate which is included in the `v0.11.0` release. Unlike `nl80211`, the `mac80211_hwsim` generic Netlink constants are only defined in the kernel module itself, so definitions are added manually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ byteorder = "1.3"
 derive_builder = "0.20"
 getset = "0.1.2"
 libc = "0.2.174"
+linux-raw-sys = { version = "0.11.0", features = ["netlink"] }
 log = "0.4"
 
 [dependencies.neli-proc-macros]
--- a/examples/mac80211_hwsim.rs
+++ b/examples/mac80211_hwsim.rs
@@ -1,0 +1,92 @@
+use std::error::Error;
+
+use neli::attr::Attribute;
+use neli::consts::mac80211_hwsim::{Mac80211HwsimAttr, Mac80211HwsimCmd};
+use neli::router::synchronous::NlRouter;
+use neli::{
+    consts::{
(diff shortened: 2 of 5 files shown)


jbaublitz (changes requested):
Hi @a-gavin, thanks for the contribution.

I have a few notes about this PR. First of all, can you provide more information on where the constants are defined for the hardcoded constants? Are they not included in the crate that you pulled in?

Overall, I'll do a final review and probably include this as soon as you fix the documentation bug popping up in CI.

PR #307, opened 2025-10-13
FromBytes family cleanups

oherrala:
Cleanups to basic types implementing `FromBytes`, `FromBytesWithInput`, `FromBytesWithInputBorrowed`.

 * Fix `&str` implementation of `FromBytesWithInputBorrowed` to handle string as null terminated. This might have been a bug? Now it also validates proper null termination without interior null bytes.
 * Change aforementioned `&str` reader to use `&[u8]`'s `FromBytesWithInputBorrowed` to access buffer
 * Change `String` reader to use `&str`'s `FromBytesWithInputBorrowed`
 * Minor touch up to `Vec<T>`'s reader for consistent style
 * Change methods `input` parameter to consistently be `Self::Input` since some were `usize`
 * Add more tests

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,12 +79,12 @@
 //!                 .build()?
 //!         ),
 //!     )?;
-//!     
+//!
 //!     for msg in recv {
 //!         let msg = msg?;
 //!         // Do things with response here...
 //!     }
-//!     
+//!
 //!     Ok(())
 //! }
 //! ```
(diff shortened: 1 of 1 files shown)


jbaublitz:
@oherrala I believe this is addressing the problem of the null terminator being included in the String, is that correct? If so, I believe this does not qualify as a bug fix and needs to target v0.8.0 unfortunately. It's not causing an error or failures and I don't want to break anyone who may have already worked around this unfortunate bug in the code.

jbaublitz:
Okay I've pulled the commits into #294 !

jbaublitz:
Thanks for the contribution.

PR #310, opened 2025-11-04
Add support for FIB rule messages

alexsikomas:
This PR adds support for managing the routing policy database through FIB rule messages, enabling the creation, deletion, and querying of ip rules.

Changes:
- `Fibmsg` struct: A message struct that corresponds to the `fib_rule_hdr` in the Linux kernel.
- `Frattr` enum: Netlink attributes for a FIB rule.
- `FrAct` enum: Possible actions for a FIB rule.
- `Frf` flags: Possible flags associated with a FIB rule
- `fwmark.rs` example: An example is included to demonstrate how to create and delete a firewall mark-based routing rule. It shows how to build a `Fibmsg`, send it to the kernel, handle the acknowledgment, and then send a corresponding delete request.

The new functionality can be tested by running the example:
```bash
# Requires network capabilities or root privileges
sudo cargo run --example fwmark
```
or
```bash
cargo build --example fwmark
sudo ./target/debug/examples/fwmark
```

When run, the example will:

1. Create a new rule: `[PRIORITY]: not from all fwmark 0xca6c lookup 246813579`
2. Print a success message.
3. Delete the rule.

You can comment out the deletion part in examples/fwmark.rs and verify the rule's creation with `ip rule show`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ byteorder = "1.3"
 derive_builder = "0.20"
 getset = "0.1.2"
 libc = "0.2.174"
+linux-raw-sys = {version = "0.12.1", features = ["netlink"]}
 log = "0.4"
 
 [dependencies.neli-proc-macros]
--- a/examples/fwmark.rs
+++ b/examples/fwmark.rs
@@ -1,0 +1,126 @@
+use neli::{
+    consts::{
+        nl::{NlmF, Nlmsg},
+        rtnl::{Frattr, Frf, RtAddrFamily, Rtm},
+        socket::NlFamily,
+    },
+    err::Nlmsgerr,
+    nl::{NlPayload, NlmsghdrBuilder},
+    rtnl::{FibmsgBuilder, RtattrBuilder},
+    socket::synchronous::NlSocketHandle,
+    types::RtBuffer,
+    utils::Groups,
+};
+
+/// Must either have network permissions (setcap) or run as sudo
+///
+/// Will make the following rule:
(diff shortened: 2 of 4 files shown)


jbaublitz:
Hi @alexsikomas, thanks for the contribution. Can you provide more information on where these constants are defined?

alexsikomas:
Hi, the definitions for these constants can be found here: [include/uapi/linux/fib_rules.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/fib_rules.h)

jbaublitz:
@alexsikomas Thanks for providing the link to the constants. I wanted to check if it's covered in linux-sys-raw. It appears that it is not currently. If you could submit a PR to linux-sys-raw or libc to have the constants automatically generated from there, I would be willing to accept this PR. As it stands, hardcoding constants often causes breakage, so I'd prefer to have it generated from C headers by another crate.

alexsikomas:
I've changed the hardcoded constants to rely on `linux-raw-sys`, which has support for `fib-rules.h` as of 0.12.1.

alexsikomas:
I've made the requested changes, removing the wildcard import, and rebased onto the current branch.

jbaublitz:
Okay I found the definition. The naming was a little bit off so it was not easy to immediately find but I believe this is fine. I'm going to merge this now, but if you can follow up and explain your usage of libc::u_char instead of u8 as it is specified in the definition, that would be helpful. If I don't hear from you, I may change the type, but based on the C spec, they're equivalent anyway so it shouldn't be a problem.

PR #330, opened 2026-06-21
Update nl80211 constants

a-gavin:
First commit is a bit pedantic and technically an API change (to a commit merged three days ago and not yet in a release). Happy to drop if requested


jbaublitz (commented):
I'm fine with the constant name changes. I think next steps, as you mentioned, are fixing the doc tests and then I'll go over the documentation.

jbaublitz (changes requested):
So far this looks fine. I don't have super strong feelings about how you name the constants as long as it's consistent and relatively close to the naming scheme used everywhere else.

Just be aware that I think some of your renaming caused breakage in the examples so you should probably take a look at that.

